PR #9, opened 2022-05-27
PHP: Add our own wrapper to Woorelease, enabling more features.

JPry:

This pull request add the first PHP code to the grow repo. This started off (very long ago 😰) as a way to install custom commands to the `~/.woorelease/extensions` directory. While I had initially hoped to treat this as a custom Composer plugin, I realized in the end that this wasn't the best approach.

While working recently on compatibility testing, and knowing that we needed to do a bunch of releases with these updates, I thought that it would be useful to have the option of doing a bulk release process. I'm sure there's more work to be done to make this nicer, but for now I've got it working in a simple way.

This PR now provides a wrapper and replacement for the Woorelease script. Rather than calling the `wr` script directly to trigger a release, we could instead use this script. 

### Notable changes from Woorelease

1. We've replaced the `vb:replace` command with our own variation. This serves the original purpose of doing replacements in more places for our needs.
2. To facilitate the above point, we have our own `\Automattic\WooCommerce\Grow\WR\Utils\VersionReplace` class that replaces `\WR\Tools\Version_Replace` in our own command.
3. We now have a `bulk` command, which is the largest change.

#### Running regular commands

The `bin/wr` script should be the new method for running Woorelease, regardless of what command is being used. This is meant to be a drop-in replacement that we can continue to tweak as needed, and I think it would ultimately be beneficial to test things out here before trying to incorporate them upstream.

#### The `bulk` command

This new command serves to automatically run either the `simulate` or `release` command against multiple extensions in succession. There are a few key components to how this is designed. There's certainly room for enhancing how this works in the future, but for now I was more interested in getting something working rather than trying to build it out as thoroughly as we'd like. I'll outline the key components first, then describe how to make use of the command.

First, all of our extensions have been defined in the [`extensions.json` file](https://github.com/woocommerce/grow/pull/9/files#diff-67cc6b3a3e8789d9c898df808bc20c696365e6356a7ecc49d28cdb7e5b43a566). Each extension has a human-friendly name, the slug for the GitHub repo, and optionally the default branch to use for releases. When the default branch isn't specified within the release data, it will use the top-level `defaultBranch` property. This file also specifies the default GitHub organization, and includes our retired extensions. The idea was to provide the data we need while also making it easily future proof.

The next important file is the [`release.txt` file](https://github.com/woocommerce/grow/pull/9/files#diff-da6108b0aaa39a3fca2d90d4fbddc241cd3fd80c0a59573754cfe74813ad1bbe). This has been added to Git, but it is intended for changes to be ignored. This file is used by the `bulk` command to determine which extensions need to be released, what version should be released, and optionally what branch should be used for release. The intention with the `release.txt` file is to copy/paste the extnsion and version columns directly from the extensions release spreadsheet. You can then optionally add a branch name as a third column to each extension.

As an example when testing this code, here's what I added to the `release.txt` file:

```
facebook-for-woocommerce	2.6.26
pinterest-for-woocommerce	1.2.4
woocommerce-points-and-rewards	1.7.17
woocommerce-brands	1.6.38
woocommerce-product-filters	1.3.8
woocommerce-coupon-campaigns	1.1.27
```

With these files in place the `bulk` command has what it needs. Here's the usage structure for the command:

```
Description:
Run a bulk simulate or release.

Usage:
bulk <release-command>

Arguments:
release-command       The command to run in bulk.

Options:
-h, --help            Display this help message
-q, --quiet           Do not output any message
-v|vv|vvv, --verbose  Increase the verbosity of messages: default is normal, 1 for verbose, 2 for more verbose output and 3 for debug
```

The only native argument for this command is `release-command`, which is either `simulate` or `release`. What's not shown with this is that all of the options from `release` or `simulate` are also valid, except for the `product_version` flag. All options passed to `bulk` will be passed to whichever command is being executed. Here's an example of the actual command:

```sh
./bin/wr bulk release --wc_tested=7.0 --generate_changelog
```

The command will then loop through all of the extensions that have been included in the `release.txt` file in sequence, separating each of them with clear output indicating the start and end of each one. This allows for easily scrolling back through the output to determine what happened. The command will also provide a list of any extensions that failed for any reason as a list at the end of the output, making it easy to identify if there were any problems without needing to keep watch or scroll back through.

1. Check out this branch
2. Run `composer install`
3. Run `./bin/wr`, and you should get output like this:

```
❯ ./bin/wr                                                  

______                      ____       __
/ ____/________ _      __   / __ \___  / /__  ____ _________
/ / __/ ___/ __ \ | /| / /  / /_/ / _ \/ / _ \/ __ `/ ___/ _ \
/ /_/ / /  / /_/ / |/ |/ /  / _, _/  __/ /  __/ /_/ (__  )  __/
\____/_/   \____/|__/|__/  /_/ |_|\___/_/\___/\__,_/____/\___/


Woorelease 2.4.0

Usage:
command [options] [arguments]

Options:
-h, --help            Display this help message
-q, --quiet           Do not output any message
-v|vv|vvv, --verbose  Increase the verbosity of messages: default is normal, 1 for verbose, 2 for more verbose output and 3 for debug

Available commands:
build              Creates a zip file.
bulk               Run a bulk simulate or release.
completion         Dump the shell completion script
detect_changes     Detect if there are changes to the product since last release.
help               Display help for a command
list               List commands
release            Creates a release, following the standard Woo process.
simulate           Simulates the release process, following the standard Woo process.
wizard             A companion tool to help you choose the correct Woorelease command.
cl
cl:generate        Creates a changelog and pushes the changes to GitHub.
gh
gh:release         Creates a GitHub release.
self
self:release       Creates a release of Woorelease, following the standard Woo process.
self:simulate      Simulates a release of Woorelease, following the standard Woo process.
translate
translate:trigger  Triggers a POT download using the translation API server.
vb
vb:change          Creates a commit that contains versions bumped.
vb:replace         Replaces placeholder PHPDoc versions. (Ventures version)
wccom
wccom:release      Creates a WCcom release.
wccom:status       Gets the WCcom status of the latest release.
wporg
wporg:release      Creates a WPorg release.
```

4. Run one of the existing Woorelease commands, such as `simulate`.
5. Modify the `release.txt` file with some content from copied from the release spreadsheet
6. Run the `bulk` command as described above with one of the commands, such as `simulate`

### Technical details

The Woorlease tool makes use of the [Symfony Console component](https://symfony.com/doc/5.4/console.html) to generate the application and the commands that it recognizes. There is then a dedicated file that loads everything, instantiates the application, and runs it.

This process allows us to make use of the same application, but to customize the load process with our own replacement process. Doing this means that we can add (or replace) commands, and also tweak the other behavior of the application.

To really delve into the differences, compare the `bin/wr` file to the `woorelease.php` file.

### Changelog entry

> Custom wrapper for the Woorelease tool, with enhanced version replacing and a bulk command


tomalec:
Addin installation instructions, base on the Slack discussion, and what I think should be adapted to here.

I think it's worth to add it to some README.md

## Installation

1. Copy `packages/php/woorelease-extension` into `~/.woorelease/`
3. `cd` into that directory, and run `composer install -o --no-dev` . As part of this process, it should create the `~/.woorelease/extensions` directory if it doesn't yet exist, and then symbolically link the `bootstrap.php` file to `~/.woorelease/extensions/bootstrap.php`.
3. Run the Woorelease command with no arguments, which should give you a list of commands. Check that the `vb:replace` command includes (Ventures version) at the end of the description.

JPry:
Apologies for the half-finished description, I accidentally closed the window I was working in before saving it. I'll get it updated tomorrow.

- [x] Update description with full details of changes and testing

budzanowski (approved):
Lets do it! Test it in production maybe 🤣

Issue #5, opened 2022-05-20
Add versioning branches or tags


puntope:
## User story
As a developer I want to refer to this repo using versioned tags instead of "trunk" branch.

#### How to reproduce the problem
For referring to this repo we should add the branch trunk.

### Describe the solution you'd like 
Use a release version like WooCommerce or other repos in do. 

## Acceptance criteria
* [ ] We should be able to load the repo using for example: https://gitpkg.now.sh/woocommerce/grow/packages/js/storybook?1.0.1

tomalec:
WDYT of using Nx (to be consistent with woocommerce/woocommerce stack) and https://github.com/jscutlery/semver plugin?
So we could simply do something like:
```sh
nx run storybook:version --releaseAs=minor
```
To do version bumps, git commit, tag (`storybook@1.2.3`) & push

puntope:
> WDYT of using Nx (to be consistent with woocommerce/woocommerce stack) 

I guess is same as NPX? Glad to use it, looks cool and I wanted to give it a go for a long time. So 👍  for it

> and https://github.com/jscutlery/semver plugin?

Never used it. (Neither heard about it) Wouldn't be enough with git flow?

tomalec:
> I guess is same as NPX? 

not really npx (at least the one I know of) is an [npm package that gives a command](https://www.npmjs.com/package/npx) to execute binaries. While Nx is a rather big tool to setup a monorepo workflows.

> Wouldn't be enough with git flow?

I don't think git flow works for monorepos out of the box. But maybe there is a way to use git flow directly with Nx. I don't know, As I've never used Nx or [`jscutlery/semver`](https://github.com/jscutlery/semver)

tomalec:
Maybe we can wait with Nx setup for https://github.com/woocommerce/woocommerce/pull/33079 as if Woo will move away to another tool, I'd use here the same.

eason9487:
Closing this issue since PR #232 has added automated release workflows for the `jsdoc` and `tracking-jsdoc` packages. All packages now have their own release processes and versioning management.

PR #11, opened 2022-06-07
Add custom GitHub actions for reuse across repositories

eason9487:

It's phase 1 of the migration plan in this comment - pb0Spc-2DL-p2#comment-5753.

Add three custom GitHub actions for reuse across repositories.

Go to the Files changed and Check tabs of the RP 1126-gh-woocommerce/automatewoo to see how these custom actions are used.

### Additional details:

About the custom composite action: https://docs.github.com/en/actions/creating-actions/creating-a-composite-action

### Changelog entry

> Add - Custom GitHub actions for reuse across repositories.

### DIFF
--- a/github-actions/prepare-mysql/README.md
+++ b/github-actions/prepare-mysql/README.md
@@ -1,0 +1,30 @@
+# Prepare MySQL in GitHub Actions
+
+This action provides the following functionality for GitHub Actions users:
+
+- Enable MySQL service
+- Handle the authentication compatibility for PHP 7.0-7.3
+- Log the version information
+
+## Usage
+
+See [action.yml](action.yml)
+
+#### Basic:
+
+```yaml
+jobs:
+  UnitTests:
+    name: Unit tests with database
+    runs-on: ubuntu-latest
+
+    steps:
+      - name: Checkout repository
+        uses: actions/checkout@v3
+
+      - name: Prepare MySQL
+        uses: woocommerce/grow/github-actions/prepare-mysql@github-actions
+
+      - name: Create database
+        run: mysql -e 'CREATE DATABASE IF NOT EXISTS wordpress_test;' -u root -proot
+```
--- a/README.md
+++ b/README.md
@@ -9,6 +9,11 @@ The packages here are too experimental or too Grow-specific to be shared Woo-wid
 - [`/packages/js/storybook`](packages/js/storybook/README.md) - Storybook dependencies and basic scripts
 - [`/packages/js/tracking-jsdoc`](packages/js/tracking-jsdoc/README.md) - `jsdoc` plugin to document Tracking Events in markdown
 
+## GitHub actions
+- [`/github-actions/prepare-mysql`](github-actions/prepare-mysql) - Enable MySQL, handle authentication compatibility
+- [`/github-actions/prepare-node`](github-actions/prepare-node) - Set up Node.js with a specific version, load npm cache, install Node dependencies
+- [`/github-actions/prepare-php`](github-actions/prepare-php) - Set up PHP with a specific version and tools, load Composer cache, install Composer dependencies
+
 <p align="center">
 	<br/><br/>
 	Made with 💜 by <a href="https://woocommerce.com/">WooCommerce</a>.<br/>
--- a/github-actions/prepare-mysql/action.yml
+++ b/github-actions/prepare-mysql/action.yml
@@ -1,0 +1,21 @@
+name: Prepare MySQL
+description: Enable MySQL, handle authentication compatibility.
+
+runs:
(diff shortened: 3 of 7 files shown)


puntope:
Howdy @eason9487 

One question before testing. I guess this PR just stores the files here and the developer should copy paste them to the projects?

I'm thinking if would be more convenient to extend the [generator](https://github.com/woocommerce/grow/tree/trunk/packages/js/generator-grow) adding some kind of GH actions generator

In my mind would be some prompt like (just an example to get the idea):

- Do you want to install GH Actions? yes/no
- Is your project using node? Yes/No
- Is your project using php? Yes/No 
- Is your project using mysql? Yes/No 

Then depending on answers the files would be generated in the folder.

tomalec:
AFAIU, individual repos do not copy those files but use a reference to here 
1126-gh-woocommerce/automatewoo/files#diff-626cf1025fab6d0ecd9b6b485d250f06348b4893aae8b6ad792ac6b409000490R55
```yml
      - name: Prepare PHP
        uses: woocommerce/grow/github-actions/prepare-php@github-actions
        with:
          php-version: "${{ matrix.php }}"
          install-deps: "no"
```

I like the idea of having a generator for workflows as well, but I'd keep actions here if they are totally same for all the repos.

:nail_care: Then we could create another PR to add workflow generator, that would setup more composite, repo-specific workflows, like 1126-gh-woocommerce/automatewoo/files#diff-626cf1025fab6d0ecd9b6b485d250f06348b4893aae8b6ad792ac6b409000490
- PHP Coding Standards
- PHP Unit tests
- (TBC) PR autolabeler

github-actions/prepare-mysql/README.md
```diff
@@ -0,0 +1,30 @@
...
+  UnitTests:
+    name: Unit tests with database
+    runs-on: ubuntu-latest
+
+    steps:
+      - name: Checkout repository
+        uses: actions/checkout@v3
+
+      - name: Prepare MySQL
+        uses: woocommerce/grow/github-actions/prepare-mysql@github-actions
```
tomalec, line 26: Maybe it would be safer to use commit hash, to avoid problems like https://github.com/woocommerce/google-listings-and-ads/pull/1552, especially that this branch may get removed after this PR is merged.

However, branch names are much easier to read :confused: 

Maybe change to `trunk`?

Seems we need to solve https://github.com/woocommerce/grow/issues/5 sooner than later

eason9487:
@puntope 

> One question before testing. I guess this PR just stores the files here and the developer should copy paste them to the projects?

Sorry, I missed mentioning the context of this PR. It's phase 1 of the migration plan in this comment - pb0Spc-2DL-p2#comment-5753.

These actions would not be coped to caller repos but only be used like [these examples](https://github.com/woocommerce/grow/blob/1dc0e5b03f160b9acce1fd1a0e3b8982c6995b15/github-actions/prepare-php/README.md#usage). The way it is actually used can be found in this AW PR - 1126-gh-woocommerce/automatewoo.

eason9487:
@tomalec and cc @puntope 

> I like the idea of having a generator for workflows as well, but I'd keep actions here if they are totally same for all the repos.

For the current scope of migrating Travis CI to GitHub Actions, I think only the Coding Standards would be completely the same or at least highly repetitive. The PHP Unit tests workflows like AW, AW-Referrals, and GLA are roughly similar but much more details are different depending on each repo.

> 💅 Then we could create another PR to add workflow generator, that would setup more composite, repo-specific workflows, [...]

I planned to complete the Travis CI to GitHub Actions migration first, and then we could have a more clear picture of what workflows have the better fit to have generators.

github-actions/prepare-mysql/README.md
eason9487: Phases 4 and 5 in pb0Spc-2DL-p2#comment-5753 are creating the versioning management for the GitHub actions in this repo. With the versioning feature, the caller repos will use `@v1-actions` to apply actions instead of a specific commit or a branch.

The `@v1-actions` will keep pointing to the latest released `v1.x.x-actions` tag, Thus, caller repos will always have the latest actions by using the `@v1-actions` major version.

I have completed the development of versioning management in the `add/versioning-action` branch and tested it with a demo release. In the PR 139-gh-woocommerce/automatewoo-referrals, we can find that the uses of `uses: woocommerce/grow/github-actions/prepare-php@v1-actions-pre` work well.

All the uses in all caller repos will be changed to `uses: woocommerce/grow/github-actions/<action-directory>@v1-actions` once the feature is formally released. I will create a separate PR for the versioning management later.

tomalec (approved):
Reviewed the code, and action logs, LGTM

PR #12, opened 2022-06-08
Add GitHub action and workflow for the versioning management of the reusable custom actions

eason9487:

It's phase 4 of the migration plan in this comment - pb0Spc-2DL-p2#comment-5753.

- Create the github-actions JS package for the [GitHub JavaScript actions](https://docs.github.com/en/actions/creating-actions/creating-a-javascript-action).
- Add a new GitHub JavaScript action for versioning management.
   - [packages/js/github-actions/actions/update-version-tags](https://github.com/woocommerce/grow/tree/762083b0fd0195f925b67cf60463f302e7851ce9/packages/js/github-actions/actions/update-version-tags)
- Add a new GitHub workflow to build bundle and manage version tags of this repo when releasing the reusable custom actions.
   - [.github/workflows/github-actions-release.yml](https://github.com/woocommerce/grow/blob/762083b0fd0195f925b67cf60463f302e7851ce9/.github/workflows/github-actions-release.yml)

### Screenshots:

After releasing the `actions-v1.0.1-pre`, both references of `actions-v1-pre` and `actions-v1.0-pre` were updated onto the same reference of `actions-v1.0.1-pre`.

1. Go to the [release workflow](https://github.com/woocommerce/grow/actions/workflows/github-actions-release.yml) to review the release processing.
1. Go to the [1557721 commit](https://github.com/woocommerce/grow/commit/155772123dad4f5f2f0eb7d85d53bac50411c474) and the [1557721 revision](https://github.com/woocommerce/grow/tree/155772123dad4f5f2f0eb7d85d53bac50411c474) to view the released version `actions-v1.0.1-pre` built and committed via the release workflow.
1. Go to the [Releases](https://github.com/woocommerce/grow/releases) and [Tags](https://github.com/woocommerce/grow/tags) pages to check the results of versioning management.
1. Go to the Files changed and Check tabs of the RP 139-gh-woocommerce/automatewoo-referrals to see how the version tag `@actions-v1-pre` is used.

### Additional details:

About the versioning management, a more detailed description is documented [here](https://github.com/woocommerce/grow/tree/05fe94cbf256d43eee5ec920238446c3a5792842/packages/js/github-actions#official-release).

### Changelog entry

> Add - A new GitHub JavaScript action for the versioning management.
> Add - A new GitHub workflow for the bundle building and versioning management of the reusable custom actions in the repository when releasing.


tomalec:
:pray: Thanks for awesome work! :clap: 

:question: I'm a bit worried about the version pattern. I think it would be nice to use it for other packages here. `1.2.3-actions` is a valid [semver](https://semver.org/), but one may try to compare `1.2.3-actions > 1.2.1-storybook` which should not take place as they are unrelated.
Have you considered something similar to [what Gutenberg does](https://github.com/WordPress/gutenberg/tags) `@wordpress/url@3.11.0`. Make a prefix instead of the suffix, to make it more clearly non-comperable.

:question: I'm still a bit confused by the different locations `/github-actions/update-version-tags/` vs `/packages/js/github-actions/update-version-tags/`. If I understood correctly, `/packages/js/github-actions/update-version-tags/` is a source that is build to `/github-actions/update-version-tags/`, and the later is something considered a published thing for external use.

While for the other tools `/packages/*/*` is the published thing.

eason9487:
@tomalec, thanks for raising these points! That were actually the things I struggled with the most during the design stage.

### Context

The value syntax of `jobs.<job_id>.steps[*].uses` in the GitHub workflow is `{owner}/{repo}/{path-to-action-dir}@{ref}`, which means it uses `@` as the separator of action's source and ref.

### Directory structure

> I'm still a bit confused by the different locations `/github-actions/update-version-tags/` vs `/packages/js/github-actions/update-version-tags/`. If I understood correctly, [...]

Yes, they are indeed what you understand them to be. Currently, we have two types of custom actions: [JavaScript action](https://docs.github.com/en/actions/creating-actions/creating-a-javascript-action) and [Composite action](https://docs.github.com/en/actions/creating-actions/creating-a-composite-action). Composite actions can be just a single action.yml file to finish the task, and further have more scripts called via the shell or even use other custome actions. JavaScript actions only have an entry point to a JS file to do the all needs for the task within it, and cannot use shell via the action.yml file.

So the purpose of `/packages/js/github-actions/` directory is the source to build scripts that be called by those custome actions. When releasing, scripts in `/packages/js/github-actions/` like `update-version-tags` will be built to the corresponding directory like `/github-actions/update-version-tags/`, and the `/github-actions/` will be the entry points for the external use.

Originally and ideally, I thought about this structure:

```
/packages/js/github-actions/
├── actions/                  # Entry points for external use
│   ├── prepare-mysql/
│   │   ├── action.yml
│   │   └── README.md
│   ├── prepare-node/         # Composite action entry point
│   │   ├── action.yml
│   │   └── README.md         # How to use this action
│   ├── prepare-php/
│   │   ├── action.yml
│   │   └── README.md
│   └── update-version-tags/  # JavaScript action entry point
│       ├── bin/
│       │   └── index.js      # Built file goes here
│       ├── action.yml
│       └── README.md
├── update-version-tags/      # Script source of the update-version-tags action
│   ├── index.js
│   ├── parse-version.js
│   └── repo-tool.js
├── another-script-source/
│   ├── index.js
│   └── utils.js
├── package.json              # The required dependent packages of the scripts, tests, build, and etc
└── README.md                 # The overall info about this Github actions tool
```

This could make all actions-related files in the same place, and the directory structure could be clearer. But the uses at the caller repos side would be lengthy, for example `woocommerce/grow/packages/js/github-actions/actions/prepare-php@v1-actions`. So I chose the current structure as the trade-off result.

💡 I came out a new thought based on this structure. We could `git mv ./packages/js/github-actions/actions/* ./` before committing the built results when releasing:

https://github.com/woocommerce/grow/blob/33f065f277e4b07dd13118d980f8ed882a7b2b26/.github/workflows/github-actions-release.yml#L54-L57

Therefore, the released directory structure would be:

```
/
├── prepare-mysql/
│   ├── action.yml
│   └── README.md
├── prepare-node/
│   ├── action.yml
│   └── README.md
├── prepare-php/
│   ├── action.yml
│   └── README.md
└── update-version-tags/
    ├── bin/
    │   └── index.js
    ├── action.yml
    └── README.md
```

With this, the caller repo could use a shorter and more meaningful source like `woocommerce/grow/prepare-php@v1-actions`.

### Version patterns

> Have you considered something similar to [what Gutenberg does](https://github.com/WordPress/gutenberg/tags) `@wordpress/url@3.11.0`. Make a prefix instead of the suffix, to make it more clearly non-comperable.

I thought about using the same pattern as Gutenberg, but the `uses` syntax use `@` as the separator and it doesn't allow another `@` in the syntax.

I then thought of several different patterns, such as `actions/v1.0.1`, `actions.v1.0.1`, `actions-v1.0.1` and `actions_v1.0.1` , but didn't think which one was more ideal. And I found that [woo-admin](https://github.com/woocommerce/woocommerce-admin/tags) use semver pattern like `v3.3.2` and `v3.3.2-plugin`, so I finally chose the semver pattern.

Considering the `{owner}/{repo}` parts in the syntax are already like the namespace as `@woocommerce-grow`, and this tool won't be published on npm. If other tools in this repo wouldn't be published on npm either,  perhaps we can use a version pattern like `actions-v1.0.1` or `actions/v1.0.1` without the namespace?

---

Comprehensive above, what do you think if we make all actions-related files in the same directory under the `/packages/js/`, and let the caller repos use syntax like:
1. `uses: woocommerce/grow/prepare-php@actions-v1`
2. `uses: woocommerce/grow/prepare-php@actions/v1`
3. Any other version patters?

tomalec:
Thank you for the detailed explanation :) 

----

### re: Directory structure

> So the purpose of `/packages/js/github-actions/` directory is the source to build scripts that be called by those custome actions. When releasing, scripts in `/packages/js/github-actions/` like `update-version-tags` will be built to the corresponding directory like `/github-actions/update-version-tags/`, and the `/github-actions/` will be the entry points for the external use.

If `/packages/js/github-actions/` is not something we will expose to anybody, maybe we could move it to `/source/github-actions`? However, that may be confusing as well, as we will have some source in `source` and some close to `packages` :(

> But the uses at the caller repos side would be lengthy, for example `woocommerce/grow/packages/js/github-actions/actions/prepare-php@v1-actions`

Yep, it's length. It makes me think, that maybe `package/(js|php)` was not a good idea. Categorizing, by the main language may not be worth prolonging pahts everywhere. (I'm not sure what were other reasons do tdo that in WC) Without it we could have:

```
/github-actions/
├── prepare-mysql/
│   ├── action.yml
│   └── README.md
├── prepare-node/         # Composite action entry point
│   ├── action.yml
│   └── README.md         # How to use this action
├── prepare-php/
│   ├── action.yml
│   └── README.md
└── update-version-tags/  # JavaScript action entry point
      ├── bin/
      │   └── index.js      # Built file goes here
      ├── src/                 # Script source 
      │   ├── index.js
      │   ├── parse-version.js
      │   └── repo-tool.js
      ├── action.yml
      └── README.md
```
So it could be used as `woocommerce/grow/github-actions/prepare-php@…`

-----


> :bulb: I came out a new thought based on this structure. We could git mv `./packages/js/github-actions/actions/* ./ `before committing the built results when releasing.

Neat, and that would only change that path for the `-actions` release branch, right?

I like that idea. However, we would need to describe that somewhere in the `trunk`, so people landing directly on GitHubs default branch don't get lost.

tomalec:
### re: Version patterns

My biggest concern is that humans and machines/tools may consider tags `1.2.3-actions` & `1.2.1-storybook` as a part of the same line and update `/storybook` to `1.2.3-actions` thinking it's the latest, while it could be older, or simply detached branch.

So I think prefix is safer. I don't mind much about `@`, we could use any delimiter, or `v`. Woo-admin is not archived, so I'd rather stick closer to WC or Gutenberg conventions.

I agree, that `{owner}/{repo}` are enough for a namespace.  Personally, I find `-v` a relatively short and good alternative to `@`.  `v` denotes the version, and `-` separates visually. Plus `-`(as opposed to `/`) kind of links to the prefix, rather than cut the separation hard, and makes the prefix one thing "actions version", "storybook version". (As opposed to "version within actions")

Therefore, my vote goes to:
> 1. `uses: woocommerce/grow/prepare-php@actions-v1`

eason9487:
@tomalec, thanks for the idea about directory structure.

> If `/packages/js/github-actions/` is not something we will expose to anybody, maybe we could move it to `/source/github-actions`? However, that may be confusing as well, as we will have some source in `source` and some close to `packages` :(

We could say all files in `/packages/js/github-actions/` are the source of this tool. The custom actions in the `/packages/js/github-actions/actions/` directory are the blueprint of this tool to be built with source and be exposed to the project root directory when released.

> It makes me think, that maybe `package/(js|php)` was not a good idea. Categorizing, by the main language may not be worth prolonging pahts everywhere.

When a tool is made up of composite languages, having a specific language in its directory path is indeed a bit unsuitable. What about still placing the source in `/packages/js/github-actions/` and moving them when we figure out the suitable structure? Or maybe place the source in `/packages/github-actions/` first?

> Neat, and that would only change that path for the `-actions` release branch, right?

This tool will merge to `trunk` branch and then branch off `trunk` to do the release. The `github-actions-release.yml` workflow will create a build commit. So it will only change paths in that committed revision. Other tools won't be affected in the `trunk`.

Another benefit is it will let this repo free to reorganize the directory structure when needed since the caller repos always use the `woocommerce/grow/<action-dir>` paths. No matter how the directory structure changes, just make sure the build processing moves exposed actions to the root.



> However, we would need to describe that somewhere in the `trunk`, so people landing directly on GitHubs default branch don't get lost.

Good idea! I will add some descriptions in README.

---

About the structure of `github-actions` itself, the reason for putting the custom actions in a directory (`actions`) to have one more layer is that it's easier to [manipulate the exposed parts](https://github.com/woocommerce/grow/blob/33f065f277e4b07dd13118d980f8ed882a7b2b26/.github/workflows/github-actions-release.yml#L54-L57) when releasing them. Otherwise, it would need to carefully avoid committing `node_modules` and other unneeded directories/files into the release build. That's why the original design was structured as in https://github.com/woocommerce/grow/pull/12#issuecomment-1151987532.

Based on your idea, I would like to adjust it a bit to facilitate the above release processing.

#### Option A

```
/packages/js/github-actions/
├── actions/                  # Entry points for external use
│   ├── prepare-mysql/
│   │   ├── action.yml
│   │   └── README.md
│   ├── prepare-node/         # Composite action entry point
│   │   ├── action.yml
│   │   └── README.md         # How to use this action
│   ├── prepare-php/
│   │   ├── action.yml
│   │   └── README.md
│   └── update-version-tags/  # JavaScript action entry point
│       ├── bin/              # Built files go here
│       │   └── index.js
│       ├── src/              # Script sources go here
│       │   ├── index.js
│       │   ├── parse-version.js
│       │   └── repo-tool.js
│       ├── action.yml
│       └── README.md
├── package.json              # The required dependent packages of the scripts, tests, build, and etc
└── README.md                 # The overall info about this Github actions tool
```

Pros:
- No confusion caused by duplicate directories, such as two `update-version-tags`

Cons:
- When there are utils to be shared custom actions, it seems to be a new problem to place the source code of the utils where it can still fit into this directory structure.
- The `src` directories will be included in the release build. (But it's not really a problem.)

#### Option B

The original structure mentioned in https://github.com/woocommerce/grow/pull/12#issuecomment-1151987532

Pros:
- Easier to add the source code of shared utils without struggling with the directory structure
- The release build is clean

Cons:
- It's a bit confused due to the duplicate directories

---

What do you think about these options?

eason9487:
@tomalec, thanks for your thoughts.

> My biggest concern is that humans and machines/tools may consider tags `1.2.3-actions` & `1.2.1-storybook` as a part of the same line and update `/storybook` to `1.2.3-actions` thinking it's the latest, while it could be older, or simply detached branch.

Agree. I was a bit worried using the `pre-release` part of semver as a tool name might also cause confusion. Changing it to a prefix would not only avoid the problem you mentioned, but also allow the `pre-release` to be used in a way that fits its definition.

According to https://github.com/woocommerce/grow/pull/12#issuecomment-1152394922, I have updated the version pattern in https://github.com/woocommerce/grow/pull/12/commits/8bc4bc900e87462d0dcdd70b9827424210f7aca0 and https://github.com/woocommerce/grow/pull/12/commits/41691ea8990cdf03249279957364bda62659defd. I will update the content and reference links in this PR description after changing the directory structure.

tomalec:
> When a tool is made up of composite languages, having a specific language in its directory path is indeed a bit unsuitable. What about still placing the source in `/packages/js/github-actions/` and moving them when we figure out the suitable structure? Or maybe place the source in `/packages/github-actions/` first?

Definitely, we don't need to fix the `packages/*` problem here, and it's better to have a separate issue/PR for it.
I'm okay with both. leaving it as is, to make it somewhat consistent with the current state, or changing to ``/packages/github-actions/``, as GHA are composite more YML than JS


> This tool will merge to `trunk` branch and then branch off `trunk` to do the release. […]
> 
> Another benefit is it will let this repo free to reorganize the directory structure when needed since the caller repos always use the `woocommerce/grow/<action-dir>` paths. 

That's great :)

tomalec:
##### re: structure of `github-actions`

> the reason for putting the custom actions in a directory (`actions`) to have one more layer is that it's easier to [manipulate the exposed parts](https://github.com/woocommerce/grow/blob/33f065f277e4b07dd13118d980f8ed882a7b2b26/.github/workflows/github-actions-release.yml#L54-L57) when releasing them

:+1: as with your idea to move actions up in the directory tree on the release branch. I think it's just better to keep the `actions` folder under the `github-actions`' package folder

----

Having some actions source only in `/actions/{name}` and some actions distributed between `/actions/{name}` and `/{name}` will keep me thinking "If I'm about to make a change should I do it here or there?", " Shouldn't we have README.md in `/update-version-tags/`?" So I vote for Option A.

Also, I don't quite get the problem in
> When there are utils to be shared custom actions, it seems to be a new problem to place the source code of the utils where it can still fit into this directory structure.

Is that not working or problematic:
```
/packages/js/github-actions/
├── actions/                  # Entry points for external use
│   ├── another-thingy/
│   │   └── …
│   └── update-version-tags/ 
│       └── …
├── utils/
│   └── do-something.js       # Util to be referenced as `../../utils/do-something.js`
└── package.json
```
and if so how does option B make it better?

tomalec:
:grey_question: WDYT of avoiding keeping source in `index.js` files? And if we really need an `index.js` - make it only an index of exports from the folder?

github-actions/update-version-tags/action.yml
```diff
@@ -0,0 +1,13 @@
...
+inputs:
+  repo-token:
+    description: The GITHUB_TOKEN secret.
+
+  sha:
+    description: The SHA1 of commit to set version tags. By default, uses the release branch or commit.
+
+runs:
+  using: node16
+  main: build/index.js
```
tomalec: :grey_question:  Related to https://github.com/woocommerce/grow/pull/12#issuecomment-1153858821, can we make it?
```suggestion
  main: update-version-tags.js
```

or

```
  main: bin/update-version-tags.js
```

eason9487:
@tomalec

> Also, I don't quite get the problem in
> 
> > When there are utils to be shared custom actions, it seems to be a new problem to place the source code of the utils where it can still fit into this directory structure.

I probably overthought it. 😂
 
> Is that not working or problematic: [...]

The functionality works. I just thought if there is a directory like `github-actions/actions/<custom-action>/src/`, then the path for the source code of shared utils would be like:
1. `github-actions/actions/utils/src/` or `github-actions/actions/utils/` - but it's not a action.
2. `github-actions/src/utils/`, `github-actions/utils/src/` or `github-actions/utils/` - but the utils directory won't be included in the build revision since we only move the content in `github-actions/actions/` to root. And, there are still `src` in it and the importing `../../../utils/do-something.js` in the source codes references to the out of the project root directory.

Even though it seems a bit confusing in the build revision, we could try not to include `src` in the build, or highlight it in the README to prompt viewers to find the correct source codes.

> and if so how does option B make it better?

With option B, all files within `github-actions/` are considered as source codes. The `bin` directories are not under it but only be created in the build revision and in a different directory structure, which is considered as build codes. Therefore, the path for the source code of shared utils could be `github-actions/utils/`.

tomalec:
> `github-actions/utils/` - but the utils directory won't be included in the build revision since we only move the content in `github-actions/actions/` to root

But that is also a problem for B, isn't it?

-----

Let's dissect it a bit. Currently, the problem with option A is that we have parts of source/non-"production" code attached and parts missing in the release branch, with option B - the source code is disjointed into two different dirs at two different levels.

Leaving solutions aside, the entry situation is like this:
1. We have composite, yml only actions, for which source code == production code, no build required
2. We have JS action, for which source code != production code, requires build step and possibly some shared utils
3. In the end we would like all the actions in the same, shallow folder
4. We don't care much if the original source code is included in the release branch or not, but if it is better to be complete.


-----

What if we treat all actions equal? If JS action needs a build phase, composite may want it to, it's just 1:1 copy ;)
Then conceptually:
- the build/release phase of every action is to move it from source folder (`/packages/js/github-actions/actions/*`) to `./*` - what [you proposed](https://github.com/woocommerce/grow/pull/12#issuecomment-1151987532) as an action in `.github/workflows/github-actions-release.yml` could in fact be an action in `packages/js/github-actions/package.json`
- the JS actions would have a bit extended build/release as it would have to run `@vercel/ncc` before and remove its source.

Alternatively, JS action could build to `.../{action-name}/` folder (not to `.../{action-name}/bin`, then we could move all action files but not recursively.

eason9487:
@tomalec, thanks for your idea.

> But that is also a problem for B, isn't it?

By option B, all source codes are not placed in the `actions` directory, so it would not include the source codes.

> Leaving solutions aside, the entry situation is like this: [...]

Agree with these points.

> What if we treat all actions equal? If JS action needs a build phase, composite may want it to, it's just 1:1 copy ;)
Then conceptually: [...]

Yes, I just made some tries locally. The same target could be done via adding some commands and setting a global .gitignore file in `.github/workflows/github-actions-release.yml`. Since the commands to be added would be slightly more complex and change the environment git settings, I probably leave these commands in the release workflow. This way they all serve the same purpose, to generate the build for release.

Comparing A and B, I think adding a bit more commands is better than having duplicate directories that bring confusion. I will make the adjustment with Option A.

github-actions/update-version-tags/action.yml
eason9487: Changed in https://github.com/woocommerce/grow/pull/12/commits/3cbffa88e87097a6339c50115cdc648939dc4d57.

eason9487:
Hey @tomalec, I have finished the adjustment of directory structure and add more descriptions in README. Could you help with a new round of code review?

packages/js/github-actions/README.md
```diff
@@ -0,0 +1,129 @@
...
+├── update-version-tags/
+│   ├── update-version-tags.js  # Built file
+│   ├── action.yml
+│   └── README.md
+└── README.md                   # Document prompts viewers to find the correct source code
+```
+
+The release build will be committed to a standalone point in the git tree via the release workflow, to make these custom GitHub actions have better paths and can be fetched faster.
+
+```mermaid
```
tomalec, line 61: :mermaid: :heart_eyes:  Oh my! how awesome is that!?

tomalec:
Awesome! I love how clean the released tree https://github.com/woocommerce/grow/tree/actions-v1.0.1-pre is. Also, there is absolutely nothing I could complain about the source.

Reviewed the code, run a test minor release, LGTM
(I hope I cleaned up tags correctly after my tests)

Thanks :pray: !

tomalec (approved):

PR #14, opened 2022-06-14
Add a GitHub action for adding branch type labels

tomalec:

Sets `type: *` and `changelog: *` labels to a created PR, according to a branch name convention

It uses actions-ecosystem/action-add-labels to keep it consistent with 
https://github.com/woocommerce/woocommerce/blob/trunk/.github/workflows/update-feedback-labels.yml

1. Pick a test repo
2. Follow the instructions in https://github.com/woocommerce/grow/compare/github-actions...add/label-action?expand=1#diff-ffef87c73e2a2202c94315b3bf7f301382b60802da7cb2db0e12935d936f8549

### Additional details:

1. This PR does follow the convention proposed in https://github.com/woocommerce/google-listings-and-ads/pull/1565. Therefore changes and unifies the set of branch prefixes and changelog entry types.
2. The flow proposed here, works well for regular single-package repos. However, it doesn't play well for monorepos, like this (`/grow`) one. I wonder how would we like to infer the package name to be set for [`[{package-name}] changelog: {type}`](https://github.com/woocommerce/grow/pull/18)? should we use branch names, like `actions/fix/123-slug`? Then `actions` or `github-actions` to follow the package name? Or maybe have another action to check the files modified, and set `package: storybook`, then let this action take that into consideration?
 @eason9487  

### Changelog entry

>

### DIFF
--- a/packages/js/github-actions/actions/branch-label/README.md
+++ b/packages/js/github-actions/actions/branch-label/README.md
@@ -2,6 +2,13 @@
 
 This action sets PR's `type: *` & `changelog: *` labels according to the branch name convention.
 
+Branch names starting
+- `(breaking|add|update|fix|tweak|doc)/` get `changelog: *`
+- `release/` get `changelog: none`
+- `add/` get also `type: enhancement`
+- `fix/` get also `type: bug`
+- `doc/` get also `type: documentation`
+
 ## Usage
 
 See [action.yml](action.yml)
--- a/packages/js/github-actions/actions/branch-label/README.md
+++ b/packages/js/github-actions/actions/branch-label/README.md
@@ -1,0 +1,21 @@
+# Branchname to Changelog label
+
+This action sets PR's `type: *` & `changelog: *` labels according to the branch name convention.
+
+## Usage
+
+See [action.yml](action.yml)
+
+#### Basic:
+
+```yaml
+on:
+  pull_request:
+    types: opened
+jobs:
+  SetLabels:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Set Labels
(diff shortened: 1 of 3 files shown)


eason9487:
> should we use branch names, like `actions/fix/123-slug`? Then `actions` or `github-actions` to follow the package name? Or maybe have another action to check the files modified, and set `package: storybook`, then let this action take that into consideration?

The latter one seems to work. I'd prefer it if possible, so that we could remember fewer conventions. 👍

packages/js/github-actions/actions/branch-label/README.md
```diff
@@ -0,0 +1,28 @@
...
+```yaml
+on:
+  pull_request:
+    types: opened
+jobs:
+  SetLabels:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Set Labels
+        uses: woocommerce/grow/github-actions/branch-label@add/label-action
```
eason9487: Note: this needs to be updated to `uses: woocommerce/grow/branch-label@actions-v1` before merging.

packages/js/github-actions/actions/branch-label/action.yml
```diff
@@ -0,0 +1,45 @@
+name: Branch Type to Label
+description: Set PR labels according to the branch name.
+
+runs:
+  using: composite
+  steps:
+    - uses: actions/checkout@v2
```
eason9487: This step could be omitted since `actions-ecosystem/action-add-labels` can be run without the codebase.

eason9487 (approved):
Testing well. LGTM.

Issue #15, opened 2022-06-17
Add GitHub workflows to help the release process of the `github-actions` packages in this repo


eason9487:
## User story

As a developer, it would be good if the release process could be done easier and clearer.

#### To make the new release more convenient, some automatic jobs were planned to be added to help the process:

1. 📌 Prepare a release content when a specific branch is created onto the target revision.
   - Fetch the automatically generated release notes provided by GitHub API.
   - Prepend the changelog content to *CHANGELOG.md*.
   - Bump the `version` in *package.json* and *package-lock.json*.
2. 📌 Based on the specific branch, create a release PR after a release content is finished.
   - Get the version from *package.json* to generate the PR title.
   - Add the release instruction in the description of the release PR.
3. 📌 Create a new release on GitHub after a release PR is approved.
   - Get the version from *package.json* to composite the release tag.
   - Get the respective content from *CHANGELOG.md* as the release notes.

#### With these workflows, the release process would be like:

1. When a package is ready to release, create a release branch onto `trunk`. For example, `release/actions`.
2. Check if the new changelog content and updated version are correct in the release PR.
   - If something needs to be revised,  append the changes in this release PR.
3. Approve the release PR if it's all good.
   - Check if the corresponding release workflows are run successfully. For example, the workflow of creating a new release or updating version tags.
4. Merge the release PR.

PR #16, opened 2022-06-17
Add an action and a GitHub workflow to get release notes via GitHub and prepare release content.

eason9487:

This PR implements part 1 in #15 - **📌 Prepare a release content when a specific branch is created onto the target revision.**

- Move some functions to the *utils* directory.
- Add `generateReleaseNotes` method to `RepoTool` class.
- Add a new util `getPackageSettings` for getting the settings of *package.json*.
- Add a new custom GitHub action for getting release notes via a GitHub API, and infer the next version and tag.
- Add a new GitHub workflow to prepare a new release for the GitHub actions package.
- Update docs for get-release-notes action and release process.

Before merge:
- [x] Delete `release/actions-demo-1` and `release/actions-demo-2` branches.
- [x] Clean up test branches if any.

#### Check simulations on Actions

💡 Since this workflow is not easy to test, I created two simulations to demonstrate the results. And two merged PRs (#11 and #12) were attached to the `[actions] changelog: add` label for [grouping the changelog](https://github.com/woocommerce/grow/blob/add/prepare-release-action/packages/js/github-actions/release-notes-config.yml) to match the version level.

1. The first one shows the initial release - [workflow result](https://github.com/woocommerce/grow/runs/6930979012?check_suite_focus=true)
   - For the first-time release, it won't bump the versions and will use the `version` value in package.json as the initial release version number.
      image
   - View the [changes](https://github.com/woocommerce/grow/commit/51acdb7ba72750ea763154de5e4b34569aa763a9) committed by the "Prepare New Release" workflow.
2. The second one simulates the version bumping and the second changelog prepending - [workflow result](https://github.com/woocommerce/grow/runs/6931071048?check_suite_focus=true)
   - For the subsequent release, it bumps the versions based on the `version` value in package.json, and increases the version level by matching the level keywords to the changelog content.
   - In this demo, the content has a `### New Features 🎉` heading that matches the minor level. So the next version is 1.1.0.
   - View the [changes](https://github.com/woocommerce/grow/commit/9de92985351ae294d0cd7b71dd5ce972584f587f) committed by the "Prepare New Release" workflow to check if the versions were updated.

#### Test the release preparation by creating the release branch

1. Create `release/actions` branch based on this PR. https://github.com/woocommerce/grow/tree/add/prepare-release-action
   
2. Go to the ["Prepare New Release" workflow](https://github.com/woocommerce/grow/actions/workflows/github-actions-prepare-release.yml) to see how it goes.

The `release/actions-demo-1` and `release/actions-demo-2` won't trigger the workflow. They were originally the `release/actions` branch, and I renamed them to keep the demo results.


.github/workflows/github-actions-prepare-release.yml
```diff
@@ -0,0 +1,81 @@
...
+
+      - name: Prepare release commits
+        # Use the github-actions bot account to commit.
+        # https://api.github.com/users/github-actions%5Bbot%5D
+        run: |
+          cd ./packages/js/github-actions
+          TODAY=$(date '+%Y-%m-%d')
+          NEXT_VER="${{ steps.get-notes.outputs.next-version }}"
+          CHANGELOG="${{ steps.get-notes.outputs.release-changelog }}"
+          CHANGELOG=$(echo "$CHANGELOG" | sed -E 's/\.? by @[^ ]+ in (https:\/\/github\.com\/.+)/. (\1)/')
```
tomalec, line 66: :construction: If we remove the PR authors and links, I think we should also remove the 

```md
## New Contributors
```

section. (See https://github.com/woorelease-bugs/initial-release/releases/new?tag=1.0.1 fro an example)

tomalec (commented):
Great tool, thanks! 
Reviewed the code, and inspected workflow logs - LGTM.

The only comments I have are:
- :construction: https://github.com/woocommerce/grow/pull/16/files#discussion_r902845866
- :nail_care: that maybe [we don't need that much granularity between feature/update/enhancement](https://github.com/woocommerce/google-listings-and-ads/pull/1565#issuecomment-1158844892) and we can keep only two. But, I'm fine with all three as well to keep it consistent with bigger projects. Plus we could always change it later.

.github/workflows/github-actions-prepare-release.yml
eason9487: The `outputs.release-changelog` output only contains the part of "## What's Changed" heading and its entries. So other contents won't be included. I added output logs in the get-release-notes action by https://github.com/woocommerce/grow/pull/16/commits/dd0226ad09ce3b314615b99ec9ce3660f6ec693c. It should be easier to know the actual values of outputs and to use them.

image

eason9487:
@tomalec, thanks for the review.

> * that maybe [we don't need that much granularity between feature/update/enhancement](https://github.com/woocommerce/google-listings-and-ads/pull/1565#issuecomment-1158844892) and we can keep only two. But, I'm fine with all three as well to keep it consistent with bigger projects. Plus we could always change it later.

I adjusted the config of release notes in https://github.com/woocommerce/grow/pull/16/commits/fc88a0632d01fbd82105f0feaa162eb8cbe3f2c0. By the way, after some tests, the bot accounts like `dependabot` don't have the `[bot]` suffix when setting the exclude list in the config.

Could you help with a new round of code reviews?

.github/workflows/github-actions-prepare-release.yml
tomalec: Ah, ok, now I got it.
`/## What's Changed\n([\d\D]+?)(?=\n\n)/i` catches all the things before the double line break => before 

```md

##New Constributors
```

tomalec (approved):
LGTM :) thanks

PR #17, opened 2022-06-21
Add a new custom action `phpcs-diff` to run PHPCS to the changed lines of code in a PR

eason9487:

It's a part of the progressively adding PHPCS task in this project - pb0Spc-2DL-p2.

- Add a new util `annotateByWorkflowCommand` to set annotations by workflow commands of GitHub Actions.
- Add a new custom action `phpcs-diff` to run PHPCS to the changed lines of code in a PR.

Before merging this PR:

- [x] Remove the `actions-v1-pre` release and related tags from this repo.
- [x] Change the version of prepare-php from `actions-v1-pre` to `actions-v1`.
   - https://github.com/woocommerce/grow/blob/b197a7154872906a0050fb55b036bf01ddd397ee/packages/js/github-actions/actions/phpcs-diff/action.yml#L13

Go to the Files changed and Check tabs of the RP 458-gh-woocommerce-points-and-rewards to see how this custom action is used. The following is the group of annotation commands Run in the "PHPCS to changed lines of code" job captured from that PR's checks.

image

### DIFF
--- a/packages/js/github-actions/actions/phpcs-diff/action.yml
+++ b/packages/js/github-actions/actions/phpcs-diff/action.yml
@@ -1,0 +1,32 @@
+name: PHPCS Diff
+description: Run PHPCS to the changed lines of code, set error annotations to a pull request.
+
+runs:
+  using: composite
+  steps:
+    # Checkout repository.
+    - uses: actions/checkout@v3
+      with:
+        fetch-depth: 2
+
+    # Set up PHP and tools.
+    - uses: woocommerce/grow/prepare-php@actions-v1-pre
+
+    #Log information
+    - shell: bash
+      run: |
+        vendor/bin/phpcs -i
+        vendor/bin/diffFilter -v
+
+    # Run PHPCS with diffFilter in the coverageChecker.
+    - shell: bash
+      run: |
+        vendor/bin/diffFilter --phpcsStrict <(git diff HEAD^...HEAD) <(vendor/bin/phpcs ./* -q --report=json) --report=phpcs 0 > /tmp/phpcs-diff.json
+        cd "${{ github.action_path }}"
+        node annotate-phpcs-report.js /tmp/phpcs-diff.json
+        cd -
+
+        TOTAL_ERRORS=$(jq ".totals.errors" /tmp/phpcs-diff.json)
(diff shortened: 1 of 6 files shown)


packages/js/github-actions/actions/phpcs-diff/action.yml
```diff
@@ -0,0 +1,32 @@
...
+    #Log information
+    - shell: bash
+      run: |
+        vendor/bin/phpcs -i
+        vendor/bin/diffFilter -v
+
+    # Run PHPCS with diffFilter in the coverageChecker.
+    - shell: bash
+      run: |
+        vendor/bin/diffFilter --phpcsStrict <(git diff HEAD^...HEAD) <(vendor/bin/phpcs ./* -q --report=json) --report=phpcs 0 > /tmp/phpcs-diff.json
```
tomalec, line 28: > `HEAD^...HEAD`

:question: Do I get it right, that we check here only the changes made in the last commit? not everything that was made in this PR?

Do those annotations accumulate somehow? co we would see the reports from the changes made in first commits?

Maybe with GHA we can use PR branch base ref directly?

eason9487: When creating a PR on GitHub, it will create a merge commit locally based on the `head` and `base` branches of the PR. The merge commit is referenced like `refs/pull/<number>/merge`. It's the same in GitHub Actions, when triggering a workflow to a pull request, it uses the revision of a merge commit to run the workflow as well. Therefore, `git diff HEAD^...HEAD` includes all diffs of a PR.

I intended to write a code comment on it, but I forgot. Added in https://github.com/woocommerce/grow/pull/17/commits/9f20f5e628ef1bc35a8196c7db77a31169eb4e55.

eason9487: To confirm this, I pushed two commits that each had a code change in different files to the RP 458-gh-woocommerce-points-and-rewards.

image

The annotations were shown in both files.

image

tomalec: Awesome, thanks for explanation.

tomalec (approved):
I reviewed the code, checked how the workflow runs in Points & Rewards, LGTM

PR #20, opened 2022-06-24
Add escaped outputs `release-notes-shell` and `release-changelog-shell` for the get-release-notes action

eason9487:
### Changes proposed in this Pull Request:

Using the `release-notes`or `release-changelog` outputs of get-release-notes action with double quotes in the shell may have problems. Such as there are backticks in a PR title, it would be a calling to a shell command:

```sh
LOG="${{ steps.get-notes.outputs.release-changelog }}"
```

And using single quotes in the shell may have problems when there are single quotes in a PR title:

```sh
LOG='${{ steps.get-notes.outputs.release-changelog }}'
```

This PR adds two outputs `release-notes-shell` and `release-changelog-shell`, which are escaped the single quotes, for use in the shell.

### Detailed test instructions:

See the test instructions in #21.

PR #18, opened 2022-06-22
Tweak the `update-version-tags` action and the `github-actions-release` workflow

eason9487:
### Changes proposed in this Pull Request:

- Adjust the commit wording of the release builds to make it clearer.
- The `repo-token` input in `update-version-tags` action is required. Add the missing `required: true` config for it.

### Detailed test instructions:

N/A

PR #21, opened 2022-06-24
Fix the issue that unescaped chars may break the release preparation of `github-actions` package

eason9487:

I encountered problems when releasing `actions-v1.0.0` since there are backticks ``` ` ``` in some PR titles such as #18.
I temporarily changed the PR title at that time. This PR fixes the problem with the new output added in #20.

There is no merged and unreleased PR that can be used to test. I reproduced the problem and created another demo to show this fix in a forked repo.

1. Go to the workflow runs in the forked repo to check the reproduced problem.
   - https://github.com/eason9487/grow/runs/7040144339?check_suite_focus=true#step:6:5
      
2. Go to the workflow runs in the forked repo to check the fix.
   - https://github.com/eason9487/grow/runs/7040395059?check_suite_focus=true#step:6:6
      
3. Check if the committed changelog has the same title as the released PR.
   - https://github.com/eason9487/grow/commit/2d605b4bdcad223e0998e15f8b775c59ecc150e6
      image

### DIFF
--- a/.github/workflows/github-actions-prepare-release.yml
+++ b/.github/workflows/github-actions-prepare-release.yml
@@ -60,17 +60,21 @@ jobs:
         # https://api.github.com/users/github-actions%5Bbot%5D
         run: |
           cd ./packages/js/github-actions
+
           TODAY=$(date '+%Y-%m-%d')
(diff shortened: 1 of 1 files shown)


tomalec (approved):
Reviewed the code and test run, LGTM :)

PR #26, opened 2022-07-07
Add a new generator for setting up PHPCS with WordPress coding standards

eason9487:

Add a new generator for setting up PHPCS with WordPress coding standards. It helps run PHPCS in the local development and GitHub Actions, and It has two modes:

1. Runs check for all PHP files.
2. Runs check for the changed lines of code only.

### Screenshots:

The changes of this PR https://github.com/woocommerce/woocommerce-google-analytics-integration/pull/218 were generated by this generator.

1. Git checkout the `add/generator-phpcs` branch of the grow repo.
2. Follow the [prerequisites](https://github.com/woocommerce/grow/tree/f27d216adf0712928923c0a1bf6d5019f3b5b3aa/packages/js/generator-grow#prerequisites) to set up yeoman.
3. Git checkout the `trunk` branch of [this repo](https://github.com/woocommerce/woocommerce-google-analytics-integration).
4. `cd` to the `woocommerce-google-analytics-integration` directory.
5. `yo grow:phpcs` and test different combinations of the prompts to see if the generator runs correctly.
6. Follow the test instructions in https://github.com/woocommerce/woocommerce-google-analytics-integration/pull/218 to see if the generated PHPCS setup can run locally.

### DIFF
--- a/packages/js/generator-grow/generators/phpcs/USAGE
+++ b/packages/js/generator-grow/generators/phpcs/USAGE
@@ -10,8 +10,8 @@ Description:
 
   - PHPCS config file
 
-    Create a phpcs.xml.dist or phpcs.xml file for configuring PHPCS.
-    It will also ask about the text domain and exclusion patterns.
+    Create a phpcs.xml.dist file for configuring PHPCS. It will also ask about
+    the text domain and exclusion patterns.
 
   - Local scripts
 
--- a/packages/js/generator-grow/generators/phpcs/USAGE
+++ b/packages/js/generator-grow/generators/phpcs/USAGE
@@ -1,0 +1,7 @@
+Description:
+  Set up PHPCS with WordPress coding standards to run the PHPCS check in the
+  local environment and GitHub Actions.
+
+  If it's introducing PHPCS into a legacy repository for the first time,
+  choosing to perform the check on the changed lines of code would be helpful
+  to fix the existing codebase gradually.
--- a/packages/js/generator-grow/generators/phpcs/USAGE
+++ b/packages/js/generator-grow/generators/phpcs/USAGE
@@ -1,7 +1,34 @@
 Description:
-  Set up PHPCS with WordPress coding standards to run the PHPCS check in the
-  local environment and GitHub Actions.
+  - Check on the changed lines of code only - Diff mode
 
-  If it's introducing PHPCS into a legacy repository for the first time,
-  choosing to perform the check on the changed lines of code would be helpful
-  to fix the existing codebase gradually.
+    Set up PHPCS with WordPress coding standards to run the PHPCS check in the
+    local environment and GitHub Actions.
+
+    If it's introducing PHPCS into a legacy repository for the first time,
+    choosing to perform the check on the changed lines of code would be helpful
+    to fix the existing codebase gradually.
+
+  - PHPCS config file
+
+    Create a phpcs.xml.dist or phpcs.xml file for configuring PHPCS.
+    It will also ask about the text domain and exclusion patterns.
+
+  - Local scripts
+
+    Add `lint:php` script into package.json.
+
+    With diff mode, it will also add `lint:php:diff` script into package.json and
+    create a bash script under ./bin to run the check locally.
+
+  - Github Actions check
+
+    Set up a workflow file under .github/workflows to check PHPCS in Github Actions.
+    It will also ask preferred PHP version of the check job in Github Actions.
+
+    With non-diff mode, the branches are used to specify the triggering branches of
+    a push event in GitHub Actions.
+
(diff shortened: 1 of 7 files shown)


puntope:
Working ✅ 

- Which PHPCS config file name is preferred? ✅  Both options tested, working good
- What are the exclusion patterns of PHPCS config to be set? ✅  Tested adding '.idea' into vim and it appeared in the config section as expected
- What is the text domain for the i18n strings of this plugin? ✅  I tested using 'test' and also the default value. They appeared correctly in the config text_domain section
- Whether to run the check in GitHub Actions? ✅  Tested No and Yes. Worked as expected
- If you would like to install the required composer packages together, please select the wanted packages ✅  Tested with different combos, worked for me.
-  What branches of push events should trigger this check in GitHub Actions? (separate branches by commas) ✅  Tested with default, adding "dev", and adding multiple branches
- Would you like to perform the PHPCS check on the changed lines of code only? ✅  Yes creates -diff workflow, No, loads the no diff version.
- What PHP version to use in GitHub Actions? ✅  tested with 7.4
- Allows to override "conflicts" ✅ 
 - Whether to set up npm scripts for running the check locally? ✅  Creates package.json script correctly (including bin/phpcs-diff.sh in case diff is selected)

puntope:
❓  One thing I noticed is that if you select 

Would you like to perform the PHPCS check on the changed lines of code only? Yes

Then it doesn't appear this question:

What branches of push events should trigger this check in GitHub Actions? (separate branches by commas)

puntope:
💅  What are the exclusion patterns of PHPCS config to be set?

WDTY of changing this to:

Would you want to see or modify the the exclusion patterns of PHPCS config? (Y/N.  Yes will launch your preferred editor)

That way we can avoid to open all the time the editor.

packages/js/generator-grow/generators/phpcs/USAGE
```diff
@@ -0,0 +1,7 @@
+Description:
```
puntope, line 1: 💅  WDYT of adding more explanations about the different options???

eason9487: Added in https://github.com/woocommerce/grow/pull/26/commits/00aa70dcb326333a860550cab51eced9c5d77f77.

eason9487:
@puntope, thanks for the review.

> ❓ One thing I noticed is that if you select
> 
> Would you like to perform the PHPCS check on the changed lines of code only? Yes
> 
> Then it doesn't appear this question:
> 
> What branches of push events should trigger this check in GitHub Actions? (separate branches by commas)

The GitHub action of checking on the changed lines of code is only triggered on pull request events but not push events. Since all pull requests are assumed to run the check no matter the branches, it doesn't need to set up branches.

---

> 💅 What are the exclusion patterns of PHPCS config to be set?
> 
> WDTY of changing this to:
> 
> Would you want to see or modify the the exclusion patterns of PHPCS config? (Y/N. Yes will launch your preferred editor)
> 
> That way we can avoid to open all the time the editor.

`yeoman` uses `Inquirer.js` to do the interactive questions in CLI, I checked both docs but didn't find a way to achieve the way you mentioned. After changing that question type to `confirm`, it still needs another `editor` type question to prompt users to enter the editor. And I don't see a reason to let users skip this step, so probably it would be kept as it is.

packages/js/generator-grow/generators/phpcs/index.js
```diff
@@ -0,0 +1,247 @@
...
+
+		const furtherQuestions = [];
+
+		// PHPCS configuration
+		if ( this.answers.installConfig ) {
+			const filenameAnswer = await this.prompt( {
+				type: 'list',
+				name: 'configFilename',
+				message: 'Which PHPCS config file name is preferred?',
+				...this._prepareConfigFileOptions(),
```
JPry: I don't think we should offer this choice. The best practice for PHPCS is to use `phpcs.xml.dist` in the Git repo.

puntope (approved):
✅  LGTM. Thanks for improving the docs and clarifying my questions.

💅  Some suggestion from @JPry to be addressed. But approving it in advance from my side.

packages/js/generator-grow/generators/phpcs/index.js
eason9487: Thanks for the suggestion. Changed in https://github.com/woocommerce/grow/pull/26/commits/6e9fef662076f5b6780cc2dc3e7a9748e4e3f918.

eason9487:
@JPry, thanks for the review. I have changed to use `phpcs.xml.dist` filename always as per your suggestion. Could you help with another round of code reviews?

JPry (approved):

PR #27, opened 2022-07-08
Fix the JSON syntax error of running `phpcd-diff` action when no changes are made

eason9487:

Fix the JSON syntax error of running `phpcd-diff` action when no changes are made. 

For some reason, if runs `phpcs-diff` action without any file changes, it will get a JSON syntax error because the `vendor/bin/diffFilter` outputs an empty result.

image

1. Create a local bash file at the repo's root directory temporality to verify the changes of this PR.
   ```bash
   #!/usr/bin/env bash
   JSON_REPORT="/tmp/phpcs-diff.json"

   # The original command
   # vendor/bin/diffFilter --phpcsStrict <(git diff HEAD^...HEAD) <(vendor/bin/phpcs ./* -q --report=json) --report=phpcs 0 > "$JSON_REPORT"

   # Simulate the result of no changes - an empty file.
   printf '' > "$JSON_REPORT"

   # Simulate the result of no PHPCS errors.
   # printf '{"files":[],"totals":{"errors":0,"fixable":0,"warnings":0}}' > "$JSON_REPORT"

   # Simulate the result of a PHPCS error.
   # printf '{"files":{"woocommerce-google-analytics-integration.php":{"messages":[{"message":"Expected 1 spaces before closing parenthesis; 0 found","source":"diffFilter","severity":1,"type":"ERROR","line":23,"column":1,"fixable":"false"}],"errors":1,"warnings":0}},"totals":{"errors":1,"fixable":0,"warnings":0}}' > "$JSON_REPORT"

   if [ ! -s "$JSON_REPORT" ]; then
     echo "No changes."
     exit 0
   fi

   node packages/js/github-actions/actions/phpcs-diff/src/annotate-phpcs-report.js "$JSON_REPORT"
   ```
2. Run the verification bash file with different diff results to see if the fix works well.

### DIFF
--- a/packages/js/github-actions/actions/phpcs-diff/action.yml
+++ b/packages/js/github-actions/actions/phpcs-diff/action.yml
@@ -28,6 +28,8 @@ runs:
         JSON_REPORT="/tmp/phpcs-diff.json"
         vendor/bin/diffFilter --phpcsStrict <(git diff HEAD^...HEAD) <(vendor/bin/phpcs ./* -q --report=json) --report=phpcs 0 > "$JSON_REPORT"
 
+        # It's a workaround that prevents the empty result from being passed as a JSON file to annotate-phpcs-report.js
+        # Related issue: https://github.com/exussum12/coverageChecker/issues/72
         if [ ! -s "$JSON_REPORT" ]; then
           echo "No changes."
           exit 0
--- a/packages/js/github-actions/actions/phpcs-diff/action.yml
+++ b/packages/js/github-actions/actions/phpcs-diff/action.yml
@@ -25,12 +25,19 @@ runs:
       # a workflow to a pull request, it uses the revision of a merge commit to run the workflow
       # as well. Therefore, `git diff HEAD^...HEAD` includes all diffs of a PR.
       run: |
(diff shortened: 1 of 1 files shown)


tomalec:
To me it looks like a bug in `diffFilter` (https://github.com/exussum12/coverageChecker/issues/72), as the line https://github.com/exussum12/coverageChecker/blob/495c9cb512cdb057cd2c63e6bda0efee4838bb3a/src/functions.php#L94-L95 logs an error, but does not throw, and returns with success `0` code.

IMO, it should either return 1 if the error occurred, and it cannot proper the result in a compliant format, or return the result in the compliant format (JSON)

tomalec (approved):
Tested locally LGTM.
:nail_care: We may add a link to the issue in the code comment, to maybe someday be able to remove this workaround :) 
(the repo is not very active, but there is always hope;) )

eason9487:
@tomalec, thanks for reporting the issue!

> We may add a link to the issue in the code comment, to maybe someday be able to remove this workaround :)

Sounds good! I added in https://github.com/woocommerce/grow/pull/27/commits/7d69c1edd9170a32f4edbcb887ce6a5a78ab1523.

PR #30, opened 2022-07-13
Fix a bash script syntax error in the `phpcs-diff` action

eason9487:

Fix the missing `)` in the action.yml of `phpcs-diff` action.

Check if the syntax of packages/js/github-actions/actions/phpcs-diff/action.yml is correct.


ianlin (approved):
LGTM 👍

PR #35, opened 2022-07-13
Add annotation actions of `eslint` and `stylelint` for annotating the linting results via their formatter

eason9487:

Add annotation actions of `eslint` and `stylelint` for annotating the linting results via their formatters.

1. View the draft PR #36 its Check tab for `eslint-annotation` action with eslint v8.x.
2. View the draft PR https://github.com/woocommerce/google-listings-and-ads/pull/1594 and its Check tab for `eslint-annotation` and `stylelint-annotation` actions with `eslint` v7.x.
   - [JavaScript linting](https://github.com/woocommerce/google-listings-and-ads/runs/7320548566?check_suite_focus=true)
   - [SCSS linting](https://github.com/woocommerce/google-listings-and-ads/runs/7320548227?check_suite_focus=true)

### DIFF
--- a/packages/js/github-actions/actions/eslint-annotation/README.md
+++ b/packages/js/github-actions/actions/eslint-annotation/README.md
@@ -1,0 +1,38 @@
+# ESLint formatter to annotate the linting results in GitHub Actions
+
+This action provides the following functionality for GitHub Actions users:
+
+- Annotate the errors and warnings of eslint results in GitHub Actions.
+
+## Usage
+
+See [action.yml](action.yml)
+
+## Compatibility
+
+- `eslint` >= 7
+
+#### Basic:
+
+```yaml
+name: ESLint
+
+on:
+  pull_request:
+
+jobs:
+  eslint:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v3
+      - uses: actions/setup-node@v3
+        with:
+          node-version: 16
+      - run: npm ci
+      - uses: woocommerce/grow/eslint-annotation@actions-v1
+        with:
+          formatter-dest: "./my-formatter.cjs"
+      - run: eslint --format ./my-formatter.cjs src
+```
+
+Uses .cjs to ensure the formatter script will be imported as a CommonJS module.
--- a/packages/js/github-actions/actions/eslint-annotation/action.yml
+++ b/packages/js/github-actions/actions/eslint-annotation/action.yml
@@ -1,0 +1,18 @@
+name: ESlint annotattion formatter
+description: Annotate eslint results via eslint formatter.
+
+inputs:
+  formatter-dest:
+    description: The relative path to place the eslint formatter script. Please use a destination file path within the caller repository since the script will depends on the eslint package installed in it.
+    default: "./eslintFormatter.cjs"
+
+runs:
+  using: composite
+  steps:
+    # Copy formatter script to the destination file path.
+    - shell: bash
+      run: |
+        SCRIPT_DEST="${{ inputs.formatter-dest }}"
+        mkdir -p $(dirname "$SCRIPT_DEST")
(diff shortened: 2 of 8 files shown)


packages/js/github-actions/actions/eslint-annotation/README.md
```diff
@@ -0,0 +1,38 @@
...
+    steps:
+      - uses: actions/checkout@v3
+      - uses: actions/setup-node@v3
+        with:
+          node-version: 16
+      - run: npm ci
+      - uses: woocommerce/grow/eslint-annotation@actions-v1
+        with:
+          formatter-dest: "./my-formatter.cjs"
+      - run: eslint --format ./my-formatter.cjs src
```
puntope, line 35: I don't understand well this `formatter-dest` option.

Following the code I see that we copy `eslintFormatter.cjs` file in the path  indicated in `formatter-dest` (`./my-formatter.cjs`) and then use that file for the formatting. 

Why not to call always those files `eslintFormatter.cjs` ???

packages/js/github-actions/actions/eslint-annotation/action.yml
```diff
@@ -0,0 +1,18 @@
...
+runs:
+  using: composite
+  steps:
+    # Copy formatter script to the destination file path.
+    - shell: bash
+      run: |
+        SCRIPT_DEST="${{ inputs.formatter-dest }}"
+        mkdir -p $(dirname "$SCRIPT_DEST")
+        echo '/* eslint-disable */' > "$SCRIPT_DEST"
+        cat "${{ github.action_path }}/eslintFormatter.cjs" >> "$SCRIPT_DEST"
```
puntope, line 18: Wondering if we can add also the run action here

`run: eslint --format ./my-formatter.cjs src` 

Maybe overridable or allowing to load rules for example.

puntope: In my mind would be cool to just load the action and forgot about anything else.

packages/js/github-actions/actions/eslint-annotation/README.md
eason9487: The action intends to call the GitHub annotation command via the formatter and still prints the original eslint results. To achieve this, `eslint` is not bundled in the script of this action but uses the same version of the caller repo. Therefore, the `eslintFormatter.cjs` script needs to be placed under the directory of caller repo to import its `eslint` package dynamically.

packages/js/github-actions/actions/eslint-annotation/action.yml
eason9487: The way `eslint` is performed can vary slightly between repos, so this action only provides the formatter for the callers to use. In addition, usually a caller repo still sets up `eslint` configuration locally, it eventually needs to know how to run `eslint` itself.

If this action performs `eslint`, it will need to add `eslint` options to the inputs of this action if want to retain the flexibility, so I think it would be better to provide only formatter.

packages/js/github-actions/actions/eslint-annotation/README.md
puntope: Ahh, ok, so we are copying the `eslintFormatter.cjs` (always named like that) into a file defined as `formatter-dest ` indicates. 

Is that correct? If so, why is necessary to have an option for naming that "script_dest" ??

puntope (approved):
✅  LGTM

It works as expected.

- Some discussion left about the need of "formatter-dest". But approved in advance.

packages/js/github-actions/actions/eslint-annotation/README.md
eason9487: > Is that correct?

Yes.

> If so, why is necessary to have an option for naming that "script_dest" ??

In case this package's users want to use another destination or avoid the collision of the same filename, it offers the ability for the caller repo to indicate another destination within its directory for placing the copied script.

Issue #4, opened 2022-05-20
Implement CI/CD


puntope:
## User story

As a developer I want to safely commit and ship new code to this repo in an automated way.

#### How to reproduce the problem
The repository doesn't have yet any CI/CD setup, neither linters or tests, this makes the code and the repositories using our packages unsafe and tedious to maintain. 

### Describe the solution you'd like 
Implement CI/CD with linters and tests.

## Acceptance criteria

* [x] There is a test suite for the main packages in the repo
* [x] There is a linter for PHP
* [x] There is a linter for JS
* [x] There is an automated task that runs all of those on each PR (aka, GitHub Actions)

eason9487:
With the recent merge of PR #225, I believe this follow-up issue can finally be considered resolved.

PR #34, opened 2022-07-13
Change the build tool of `github-actions` package to `rollup`

eason9487:
### Changes proposed in this Pull Request:

This PR changed the build tool of `github-actions` package to `rollup` since `@vercel/ncc` not flexible enough to meet the needs of subsequent builds for the formatters of eslint and stylelint (#35).

### Detailed test instructions:

1. View the check of https://github.com/woocommerce/grow/runs/7319890182?check_suite_focus=true and the related release PR #33 for `get-release-notes` action.
2. View the PR 1028-gh-woocommerce/woocommerce-follow-up-emails and its check tab for `phpcs-diff` action.
3. View the check of https://github.com/woocommerce/grow/runs/7319795582?check_suite_focus=true for `update-version-tags` action.

PR #36, opened 2022-07-13
Add eslint and JavaScript linting in GitHub Actions

eason9487:

Implements parts of #4 

- Add `eslint`, related deps and their configurations.
- Add a GitHub Actions workflow to run JavaScript linting.
- Fix all linting errors and warnings.

Before merging this PR:

- [x] Wait for #35 to be released.
- [x] Change `actions-v1-pre` to `actions-v1`.

1. `npm install`
2. `npm run lint:js` locally to test the eslint
   - 💡 It may gets some `import/no-unresolved` warnings because the *node_modules* directories in sub-packages don't exist. Those warnings will disappear after running `npm install` in those sub-packages.
3. View the [check](https://github.com/woocommerce/grow/runs/7320238300?check_suite_focus=true) that was intentionally left with linting errors and warnings.
4. View the annotations attached to this PR:
   github com_woocommerce_grow_pull_36_files
5. View the Check tab of this PR to see the successful linting check.


.github/workflows/js-linting.yml
```diff
@@ -0,0 +1,43 @@
...
+      - trunk
+    paths:
+      - "**.js"
+      - "**.mjs"
+      - .github/workflows/js-linting.yml
+  pull_request:
+    paths:
+      - "**.js"
+      - "**.mjs"
+      - .github/workflows/js-linting.yml
```
puntope, line 15: 💅  Maybe running this WorkFlow on PR against trunk would be enough?

.github/workflows/js-linting.yml
```diff
@@ -0,0 +1,43 @@
+name: JavaScript Linting
```
puntope, line 1: 💅  If we are going to set up a global package.json I'd also add a `.nvmrc` file

puntope:
❓ 

When I run `npm run lint:js` I got this error:

Screenshot 2022-07-15 at 16 49 55

.github/workflows/js-linting.yml
eason9487: If the linting check only runs on PR against trunk, PR to PR, like this current one (#34), wouldn't be run. I think finding out linting errors earlier should be ideal.

.github/workflows/js-linting.yml
eason9487: Added in https://github.com/woocommerce/grow/pull/36/commits/fa4b3463c1bbb7f9196e9c14d8e18b0f1f998f99.

eason9487:
@puntope, thanks for the review.

> When I run `npm run lint:js` I got this error: [...]

Thanks for catching this. Seems it results in a different node_modules structure in the initial install, and somehow the flatted structure of package-lock after running `npm dedupe` conflicts with `npm install`. I added the `"prettier": "npm:wp-prettier@2.6.2"` by https://github.com/woocommerce/grow/pull/36/commits/6489389eb8a858f39f188c12730c7c47d69471b7 to make it stable.

Could you help with another round of code reviews?

.github/workflows/js-linting.yml
puntope: What I meant is to remove the `push` directive.

puntope (approved):
✅  LGTM

- Tested `lint:js` locally. I checked forcing some errors. 
- Checked the action check
- Checked the code quality

.github/workflows/js-linting.yml
eason9487: The temporarily merged code of a PR to be run checks is not always the same as the merged one after a PR actually gets merged.

For example, if there are other commits are pushed to trunk after this PR was opened, this PR won't be triggered and run the check again if there are no conflicts. When two PR change the same file like this:

```diff
const config = {
+	hello: '123', // Added by PR A
	howdy: '000',
+	hello: '456', // Added by PR B
 };
};
```

Both PR could pass the check, but it would be a problem after merging. When this happens, the check run on push events of trunk might find out this problem.

puntope: Right! Thanks for the explanation 👌

PR #44, opened 2023-03-06
Add coverage Report as PR comment

mikkamp:

This PR adds a workflow to upload a coverage report as a PR comment.

We can see it in action on this PR: https://github.com/woocommerce/automatewoo/pull/1356

Once approved this PR can be merged and released following the instructions here: https://github.com/woocommerce/grow/tree/trunk/packages/js/github-actions
When that's completed we can switch the action `uses` source within the AutomateWoo PR.

> **Note**
This action creates a PR comment for a coverage report regardless of the language of the unit tests. So it can be used for a PHP or JS unit test report, as long as the report is in the "clover" format.

See test instuctions for PR https://github.com/woocommerce/automatewoo/pull/1356

### Changelog entry
* Dev - Add coverage report as PR comment

### DIFF
--- a/packages/js/github-actions/actions/coverage-report/README.md
+++ b/packages/js/github-actions/actions/coverage-report/README.md
@@ -1,0 +1,39 @@
+# Prepare PHP in GitHub Actions
+
+This action adds a Unit test coverage report to the PR as a comment. The coverage report must be in the clover format.
+For main branches (trunk) the coverage report will be uploaded as a GitHub artifact. When the action is run as part of a PR the artifact from the main branch will be downloaded, and use as a comparison with the report in the current PR.
+This final report will be added to the PR as a comment.
+
+## Usage
+
+See [action.yml](action.yml)
+
+By default the artifact will be saved as `coverage-report` and the report will be found in the file `tests/coverage/report.xml`. The options `report-file`, `report-path` and `report-name` can be adjusted if we need to have multiple reports (one for JS unit tests and another for PHP unit tests).
+
+The parameter workflow must point to the current workflow file, which indicates where to save the downloaded artifact so it can be used in the comparison between branches.
+
+#### Basic:
+
+```yaml
+jobs:
+  UnitTests:
+    name: Unit tests
+    runs-on: ubuntu-latest
+
+    steps:
+      - name: Checkout repository
+        uses: actions/checkout@v3
+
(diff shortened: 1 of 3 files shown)


packages/js/github-actions/actions/coverage-report/README.md
```diff
@@ -0,0 +1,39 @@
+# Prepare PHP in GitHub Actions
```
puntope: I guess this should be 

# PHP Coverage report

puntope:
One issue I see is that the Coverage action is PHP based, so I guess a better location could be in packages/actions instead of in the packages/js folder. But seems like it is not an issue introduced in this PR  and other actions have been aded in this location before

packages/js/github-actions/actions/coverage-report/action.yml
```diff
@@ -0,0 +1,69 @@
+name: Coverage Report
```
puntope, line 1: Maybe is worth it to change the defaults to include the PHP keyword as we did in AW implementation. So we can save to add them in our implementations over different plugins. 

         ```
          report-name: php-coverage-report
          report-path: tests/php-coverage
          test-comment: "PHP unit test coverage"
```

mikkamp: I originally created this PR when I added it for GLA, which does both PHP and JS unit tests. So I specifically wanted to leave it independent from the language, as it should be usable for either.

packages/js/github-actions/actions/coverage-report/README.md
mikkamp: Thanks, yes I forgot to fix that title. Although I think we should leave out the PHP, because it should be reusable for JS unit tests. The only part that refers to PHP is the example as it shows how to run it with PHP unit tests.

mikkamp:
I've update the title and added a note explaining the action can be run for both JS and PHP (so we shouldn't default to a particular language). It's ready for another round of review.

puntope (approved):

puntope:
✅  LGTM

eason9487:
Thanks for adding this action. It would be a bit nicer to also add this action to [the **Actions list** section in this README.md](https://github.com/woocommerce/grow/blob/add/coverage-report/packages/js/github-actions/README.md#actions-list).

PR #45, opened 2023-03-22
Add GitHub Action to facilitate producing documentation of WordPress hooks within a codebase

JPry:

This pull request creates a new GitHub Action for the purpose of documenting WordPress hooks within a plugin. 

For the full details of how to make use of this action, view the included [readme.md](https://github.com/woocommerce/grow/pull/45/files?short_path=b00c269#diff-b00c269d731ef3f4a451e65c2be84de70cb0cb37d4885f946b980e5c8d1bf8e8) file.

This also integrates PHP-based GitHub actions with the existing JavaScript actions.

1. [Validate that the action compiles correctly](https://github.com/woocommerce/grow/actions/runs/4889783038)
2. Test against the sample release of this action, as seen [here](https://github.com/woocommerce/woocommerce-order-source-attribution/pull/54/files#diff-2d7f014ef2360f8a72ac3dfdca4c9c52f2774fe10388f22d33303074f9ba7384)

### Changelog entry

> Add - New GitHub Action, `hook-documentation`
> Update - Add the ability to build and develop further GitHub actions using PHP.


.github/workflows/github-actions-release.yml
```diff
@@ -98,22 +98,29 @@ jobs:
...
 
-          git checkout HEAD^ -- ./packages/js/github-actions/actions
+          git checkout HEAD^ -- ./packages/js/github-actions/actions ./packages/php/github-actions
           git restore --staged .
           echo "/packages/js/github-actions/actions/*/src" > .gitignore
           git add ./packages/js/github-actions/actions
           git mv ./packages/js/github-actions/actions/* ./
+          
+          find ./packages/php/github-actions -name "*.zip" -exec unzip -d ./ {} \;
+          git add .
```
eason9487: Ideally, this process builds a release revision that can be run directly in node or php, with the required non-built-in dependencies also inlined.

However, this made the [packages](https://github.com/woocommerce/grow/tree/a14aa245eb025c2d90a0f546ca2fd4bae6698c5a/packages) directory to be committed together, and the directory includes:
- The source codes of `php/github-actions`
- The source codes of each action in `js/github-actions`
- `hook-documentation.zip` file
- `node_modules` directory
- `vendor` directory

In addition, there are a few files that were committed as well, such as the `.gitignore` in the root directory. [The contents of this link](https://github.com/woocommerce/grow/tree/actions-v1.4.1) can be a reference for the build results.

Could it still make it only commit the minimum required files?

packages/php/github-actions/hook-documentation/action.yml
```diff
@@ -0,0 +1,50 @@
...
+    default: 'no'
+
+outputs:
+  hook-docs:
+    description: The documentation generated by the script
+    value: ${{ steps.run-documentation-generator.outputs.hook-docs }}
+
+  debugging-output:
+    description: Extra debugging output from some commands
+    value: ${{ steps.debugging-output.debugging-output }}
```
eason9487: Looks like there is no content written to this output? It would be a bit better to be removed if only used in debugging.

packages/php/github-actions/hook-documentation/readme.md
```diff
@@ -0,0 +1,95 @@
...
+      - name: Checkout the repository
+        uses: actions/checkout@v3
+        with:
+          # Checks out a branch instead of a commit in detached HEAD state
+          ref: ${{ github.head_ref }}
+
+        # This generates the documentation string. The `id` property is used to reference the output in the next step.
+      - name: Generate hook documentation
+        id: generate-hook-docs
+        uses: woocommerce/grow/packages/php/github-actions/hook-documentation@feature/hook_helper
```
eason9487: Sidenote: This `uses` needs to be replaced with `woocommerce/grow/hook-documentation@actions-v1` before merging this PR.

eason9487 (commented):
Thanks for the work.

I viewed files other than PHP and will leave the code review of PHP files up to the team Ventures.

The main concern is about the build result includes a lot of unexpected files. Although this is not a mandatory constraint, it would be perfect if the build result could include the minimum required files when possible. The rest are minor comments about the action documentation.

The `github-actions` release processing of this repo will only use PRs' title as the changelog in the release note. For example, [actions-v1.3.0](https://github.com/woocommerce/grow/releases/tag/actions-v1.3.0). Maybe this PR's title could be described more explicitly.

.github/workflows/github-actions-release.yml
JPry: > Ideally, this process builds a release revision that can be run directly in node or php, with the required non-built-in dependencies also inlined.

While it's possible to consolidate down into a single file in PHP (or technically, an archive), I don't think the desire for a single file necessitates the overhead of building that. Generally speaking, PHP applications span multiple files, even in production.

> However, this made the [packages](https://github.com/woocommerce/grow/tree/a14aa245eb025c2d90a0f546ca2fd4bae6698c5a/packages) directory to be committed together, and the directory includes:
> 
> - The source codes of `php/github-actions`
> - The source codes of each action in `js/github-actions`
> - `hook-documentation.zip` file
> - `node_modules` directory
> - `vendor` directory
> 
> In addition, there are a few files that were committed as well, such as the `.gitignore` in the root directory. [The contents of this link](https://github.com/woocommerce/grow/tree/actions-v1.4.1?rgh-link-date=2023-05-12T08%3A42%3A27Z) can be a reference for the build results.
> 
> Could it still make it only commit the minimum required files?

Good catch. My goal was to build this in such a way to allow for future actions to be developed, and the release process handles them automatically. But as you point out, simply using `git add .` adds more than I intended.

(As an aside, I understand *why* we do the build this way, but I think it would be worth reconsidering how the repo is managed to facilitate a more intuitive process, and to also prevent warnings that the tag commits aren't part of the repo.)

I added 8d1189bcb09007fc8ffb1b68eea685e8f1904d9d to limit the files that are included during the commit process.

packages/php/github-actions/hook-documentation/action.yml
JPry: The idea was to allow for debugging within workflows that call this action. It only produces output when the action fails, or when debugging output is enabled.

packages/php/github-actions/hook-documentation/readme.md
JPry: Updated in bac2e77.

JPry:
@eason9487 Thanks for the review!

> The github-actions release processing of this repo will only use PRs' title as the changelog in the release note. For example, [actions-v1.3.0](https://github.com/woocommerce/grow/releases/tag/actions-v1.3.0). Maybe this PR's title could be described more explicitly.

✅

.github/workflows/github-actions-release.yml
eason9487: > While it's possible to consolidate down into a single file in PHP (or technically, an archive), I don't think the desire for a single file necessitates the overhead of building that. Generally speaking, PHP applications span multiple files, even in production.

Thanks for the clarification.

Since the npm ecosystem can easily bloat the content in `node_modules`, it's not ideal to commit the directory together for a release. Running `npm install` on the client side before performing it may take longer and have compatibility issues. Therefore, the practice of GitHub Actions written in JavaScript is usually to build them into a single file or a handful of files, and there are opportunities to reduce the overall size through tree-shaking techniques.

I agree that the PHP case is different, and guess whether or not to do further build processing is more dependent on the actual use case rather than the way in JS.

> (As an aside, I understand _why_ we do the build this way, but I think it would be worth reconsidering how the repo is managed to facilitate a more intuitive process, and to also prevent warnings that the tag commits aren't part of the repo.)

The main reason for the intuitive build process is because of the above npm ecosystem and compatibility considerations.

Another consideration is to minimize the time spent on the private repos to use these GH Actions, which in turn reduces the cost spent on GH. Therefore, the size of the release and the direct executability are also the main indicators.

Not sure about the "tag commits aren't part of the repo" warnings. Could you provide me with a practical example?

packages/php/github-actions/hook-documentation/action.yml
eason9487: Not sure if I missed anything, but I didn't find a place setting the value of `debugging-output` to `$GITHUB_OUTPUT` like the `hook-docs`. Therefore, I thought this output value is unable to be used.

https://github.com/woocommerce/grow/blob/bac2e7784a78a02a0f38d6daa61fb6777c58e893/packages/php/github-actions/hook-documentation/action.yml#L32-L34

In addition, if the output value is written from somewhere, then it seems that this line needs to be adjusted as well. Looks like it should be: `value: ${{ steps.debugging-output.outputs.debugging-output }}` ?

eason9487 (approved):
Thanks for the adjustment. 

I still viewed non-PHP files only and the PR looks good. I will be approving this PR and leaving the code review of PHP files up to the team Ventures.

.github/workflows/github-actions-release.yml
JPry: > Since the npm ecosystem can easily bloat the content in node_modules, it's not ideal to commit the directory together for a release. Running npm install on the client side before performing it may take longer and have compatibility issues. Therefore, the practice of GitHub Actions written in JavaScript is usually to build them into a single file or a handful of files, and there are opportunities to reduce the overall size through tree-shaking techniques.
> 
> I agree that the PHP case is different, and guess whether or not to do further build processing is more dependent on the actual use case rather than the way in JS.

Thanks for the explanation about the NPM process, it makes sense why you'd want to do all of the processing and consolidation ahead of time. 

With PHP's `composer` for dependency management, it is fairly efficient when set up correctly. Just like with NPM, you can define production vs. development dependencies, and composer allows for loading up dependencies that are only needed for production. In particular, the build process that's executed earlier builds a production-ready version of the dependencies, and utilizes an optimized autoloader for file inclusion.

In short, the building that's done to generate the zip files is the most efficient version of the project possible, and we're only including what's necessary for the project to run. Further, the testing I've done shows that it completes very quickly.

> Not sure about the "tag commits aren't part of the repo" warnings. Could you provide me with a practical example?

If you view the commit for any of the releases, or just browse the codebase from one of the release tags, GitHub shows a warning. Here's an example: https://github.com/woocommerce/grow/tree/actions-v1.4.1

packages/php/github-actions/hook-documentation/action.yml
JPry: > Not sure if I missed anything, but I didn't find a place setting the value of `debugging-output` to `$GITHUB_OUTPUT` like the `hook-docs`. 

🤦🏻 Good catch! 16f5618 fixes that oversight.

.github/workflows/github-actions-release.yml
eason9487: > > Not sure about the "tag commits aren't part of the repo" warnings. Could you provide me with a practical example?
> 
> If you view the commit for any of the releases, or just browse the codebase from one of the release tags, GitHub shows a warning. Here's an example: https://github.com/woocommerce/grow/tree/actions-v1.4.1

Since this repo also contains other packages and has to support multiple Custom GH actions in a single repo, (most of the Custom GH actions are one action for one repo construct), we came to the current structure after some discussion. Related discussions are in #12

To reduce confusion, the README.md below [actions-v1.4.1](https://github.com/woocommerce/grow/tree/actions-v1.4.1) briefly explains that it is a version-specific release build and guides to the actual revision of the source code. The development page also demonstrates the composition of version tagging in the git tree.

The reason for intentionally not having release commits and tags be created on a `trunk` or a specific branch is to avoid that branch's commit history looking more confusing. After all, each build removes all other packages and moves itself to the root directory, so if that branch has this particular commit and then reverts it, there may not be less complexity.

With these limitations and premises, it does result in some complexity and is not intuitive, and I would be happy to see if there is a better way to construct it and still fulfill the needs.

- Coexist with other packages in the grow repo
- Support multiple Custom GH actions in a single repo
- Released Custom GH actions are kept as thin and ready to run immediately as possible
- The caller repos can always use a path in `woocommerce/grow/<action-dir>@actions-v1` pattern

packages/php/github-actions/hook-documentation/src/Documentor.php
```diff
@@ -0,0 +1,360 @@
...
+ * Class Documentor
+ *
+ * @since x.x.x
+ */
+class Documentor {
+
+	/** @var array */
+	protected array $args;
+
+	public function __construct( array $args = [] ) {
```
rawdreeg, line 33: Maybe some doc bloc to describe the $args array?

rawdreeg (approved):
I reviewed and tested this. I didn't find anything major. Except for the small comment below

I'll approve this ✅

packages/php/github-actions/hook-documentation/src/Documentor.php
JPry: Done in ee0b6b3.

PR #50, opened 2023-04-20
[Hackday] Action Get Plugin Releases

puntope:

This PR implements an action for getting the latest plugin releases.

Screenshot 2023-04-20 at 14 52 31

Screenshot 2023-04-20 at 15 35 48

1. [Check pre-release compiles ok](https://github.com/woocommerce/grow/actions/runs/4753753435)
2. [Check test workflow is running ok ](https://github.com/woocommerce/grow/actions/runs/4753361988/jobs/8444847812)
3. [Check action in GLA running and loading properly the matrix](https://github.com/woocommerce/google-listings-and-ads/actions/runs/4753633393/jobs/8445736760)
4. [Check the workflow being added here](https://github.com/woocommerce/google-listings-and-ads/pull/1954)

### Changelog entry

> Add - Action for getting releases from WordPress or plugins


packages/js/github-actions/actions/get-plugin-releases/action.yml
```diff
@@ -0,0 +1,24 @@
...
+  includePatches:
+    description: Weather to include Patches in the fetched versions. (Wordpress is Not compatible with this)
+    default: "false"    
+    
+outputs:
+  matrix:
+    description: A matrix array in the format ["7.4","7.5","7.6"]
+
+runs:
+  using: node16
```
tomalec, line 23: :grey_question: Out of curiosity, why 16?

packages/js/github-actions/actions/get-plugin-releases/src/get-plugin-releases.js
```diff
@@ -0,0 +1,122 @@
...
+}
+
+function setOutput( key, value ) {
+	core.info( `==> Output "${ key }":\n${ value }` );
+	core.setOutput( key, value );
+}
+
+function parsePluginVersions( releases = {} ) {
+	const slug = getInput( 'slug' );
+	const numberOfReleases = parseInt( getInput( 'releases' ), 10 ) || 3;
```
tomalec: :question:  Isn't the default `3` assured by the `.yml` config? I'd rather keep the default in one place.

packages/js/github-actions/actions/get-plugin-releases/src/get-plugin-releases.js
```diff
@@ -0,0 +1,122 @@
...
+	const includeRC = getInput( 'includeRC' ) || false;
+	const includePatches = getInput( 'includePatches' ) || false;
```
tomalec: Same as [above](https://github.com/woocommerce/grow/pull/50/files#r1182780727), aren't default given by the yaml?

packages/js/github-actions/actions/get-plugin-releases/src/get-plugin-releases.js
```diff
@@ -0,0 +1,122 @@
...
+
+	const output = [];
+
+	if ( slug !== 'wordpress' ) {
+		Object.keys( releases.versions )
+			.reverse()
+			.forEach( ( version ) => {
+				if ( output.length === numberOfReleases ) {
+					return;
+				}
```
tomalec: :nail_care: I wonder if a regular `for` be cleaner to stop the loop instead of calling noop for all remaining versions.

packages/js/github-actions/actions/get-plugin-releases/src/get-plugin-releases.js
```diff
@@ -0,0 +1,122 @@
...
+
+				if (
+					version !== 'other' &&
+					version !== 'trunk' &&
+					! version.includes( 'beta' ) &&
+					! includesRC( version, includeRC ) &&
```
tomalec: The fact that for "1.2.3-rc.1" "includesRC" may return `false` is quite confusing IMHO. 
```js
( includeRC || ! includesRC( version ) )
```
would be more straightforward to me.

Also, "**includesRC**( release.version, **includeRC** )" gives me a bit of :thinking:, and I think it is pretty bug-prone (at least to my dyslectic self;) ) to miss just `s`

packages/js/github-actions/actions/get-plugin-releases/src/get-plugin-releases.js
```diff
@@ -0,0 +1,122 @@
...
+/**
+ * Internal dependencies
+ */
+import handleActionErrors from '../../../utils/handle-action-errors';
+
+async function getPluginReleases() {
+	const slug = getInput( 'slug' );
+	const apiEndpoint = getAPIEndpoint( slug );
+
+	fetch( apiEndpoint )
```
tomalec: :construction: 
```suggestion
return fetch( apiEndpoint )
```
Otherwise, there's nothing to `await` form, and we [notify finish before any work is even started](https://github.com/woocommerce/grow/actions/runs/4753361988/jobs/8444847812#step:2:10):

packages/js/github-actions/actions/get-plugin-releases/action.yml
```diff
@@ -0,0 +1,24 @@
+name: Get Plugin L-2 release versions
+description: Get release versions for the L-2 releases from a specific Wordpress plugin like WooCommerce or Google Listing And Ads.
```
tomalec:
```suggestion
description: Get release versions for the L-x releases from a specific Wordpress plugin like WooCommerce or Google Listing And Ads.
```

It's not only for "L-**2**" :)

packages/js/github-actions/actions/get-plugin-releases/README.md
```diff
@@ -0,0 +1,64 @@
+# Get latest releases versions from a specific plugin or from wordpress core
+
+This action provides the following functionality for GitHub Actions users:
+
+- Get releases versions via GitHub job
```
tomalec:
```suggestion
- Get L-x release versions via GitHub job
```

Just to emphasize the point of L-x, not all or an arbitrary number.

tomalec (commented):
Thank you for this action!
I think it will not only make our life easier by reducing the manual labor of updating the versions but also allow us to test plugins automatically in advance as the dependencies are released.

So we could have a job that periodically runs all tests against the latest betas :)


I left a few comments, also some code comments would be nice to have :)

packages/js/github-actions/actions/get-plugin-releases/action.yml
```diff
@@ -0,0 +1,24 @@
...
+
+inputs:
+  slug:
+    description: The plugin slug to get the versions.
+    required: true
+  releases:
+    description: Number of releases to get. By default L-2 directive (3 releases)
+    default: "3"
+  includeRC:
+    description: Weather to include Release Coadidates in the fetched versions. (Wordpress is Not compatible with this)
```
tomalec:
```suggestion
    description: Weather to include Release Candidates in the fetched versions. (WordPress is Not compatible with this)
```

packages/js/github-actions/actions/get-plugin-releases/action.yml
```diff
@@ -0,0 +1,24 @@
...
+    default: "false"
+  includePatches:
+    description: Weather to include Patches in the fetched versions. (Wordpress is Not compatible with this)
```
tomalec:
```suggestion
    description: Weather to include Patches in the fetched versions. (WordPress is NOT compatible with this)
```

packages/js/github-actions/actions/get-plugin-releases/README.md
```diff
@@ -0,0 +1,64 @@
...
+    name: Get Plugin Releases
+    runs-on: ubuntu-latest
+    steps:
+      - name: Get Release versions from WooCommerce
+        id: wc-matrix
+        uses: woocommerce/grow/get-plugin-releases@actions-v1.5.0-pre
+        with:
+          slug: woocommerce
+
+      - name: Get Release versions from Wordpress
```
JPry:
```suggestion
      - name: Get Release versions from WordPress
```

packages/js/github-actions/actions/get-plugin-releases/action.yml
mikkamp: Weather > Whether

mikkamp:
This looks like a well needed project, thanks for the hard work.

I just had a question about a list of versions I noticed in the screenshot:
`The 4 versions WC RC are [7.6.0-rc.3,7.5.1,7.4.1,7.3.0-rc.2].`

Picking up the newest release candidate for 7.6.0 is a good thing, but why is 7.3.0 not considered "more recent" than 7.3.0-rc.2?
Does that mean that as soon as 7.6.0 is released (and before 7.6.1 is available) 7.6.0-rc.3 will still be considered "more recent"?

packages/js/github-actions/actions/get-plugin-releases/action.yml
puntope: I was trying to be consistent with other project actions.

For example: 

https://github.com/woocommerce/grow/blob/trunk/packages/js/github-actions/actions/get-release-notes/action.yml
https://github.com/woocommerce/grow/blob/trunk/packages/js/github-actions/actions/update-version-tags/src/update-version-tags.js

puntope:
> This looks like a well needed project, thanks for the hard work.
> 
> I just had a question about a list of versions I noticed in the screenshot: `The 4 versions WC RC are [7.6.0-rc.3,7.5.1,7.4.1,7.3.0-rc.2].`
> 
> Picking up the newest release candidate for 7.6.0 is a good thing, but why is 7.3.0 not considered "more recent" than 7.3.0-rc.2? Does that mean that as soon as 7.6.0 is released (and before 7.6.1 is available) 7.6.0-rc.3 will still be considered "more recent"?

Nice catch @mikkamp , as always.

You're right. Unfortunately, the API returns the versions sorted alphabetically. 

See -> https://api.wordpress.org/plugins/info/1.0/woocommerce.json 

I created a custom sort algorithm here https://github.com/woocommerce/grow/pull/50/commits/580295b5df0163f6acf9288772343525ae90acf6

Now it should work 

see https://github.com/woocommerce/grow/actions/runs/4893825701/jobs/8737236994?pr=50

puntope:
@tomalec  I applied all your suggestions in the code. Thanks

puntope:
Howdy @tomalec @mikkamp 

I applied all the changes, this is ready for another round:

Workflow running: https://github.com/woocommerce/grow/actions/runs/4893825701/jobs/8737236994?pr=50

mikkamp:
Just curious about the test "Get L-3 Release versions from WC including RC" it's run with setting releases to 4 and it returns:
`7.7.0-rc.1,7.6.1,7.5.1,7.4.1`

But if we would have run that a day before the release candidate came out then it would be testing 7.6 - 7.3 which would have been L-4. In most our cases we are going to want to run tests for L-2 and include an extra test for a release candidate if it's available. How would we go about configuring that or do you think it's acceptable to just keep testing the old release until the release candidate becomes available?

puntope:
> 

From my understanding, L-X is a question of naming in the workflow. It's possible to call it "L-2 with RC" and set up the number of releases as 3, if that makes sense 

We are counting the RC as a Release. I guess it might be even an advantage to have it as a way to trigger potential alarms in case of error.  

Adding the config you suggest is possible. By just not counting the RCs as "releases" so setting L-2 would fetch:

7.6, 7.5, 7.4

In case an RC appears. It will fetch:

7.7.rc-1, 7.6, 7.5, 7.4

I'm ok doing both options. Just notice that doing the suggested solution might add an extra version in the matrix (the RC)

mikkamp:
L-2 is what we are required to support, but if there is a release candidate available then why not test early. So from that perspective I think it makes sense to not count the RC's.

packages/js/github-actions/actions/get-plugin-releases/src/get-plugin-releases.js
```diff
@@ -0,0 +1,141 @@
...
+	) {
+		return true;
+	}
+}
+
+function semverCompare( a, b ) {
+	const regex = /^(\d+)\.(\d+)\.(\d+)(-rc\.\d+)?$/;
+
+	const [ , majorA, minorA, patchA, rcA ] = a.toLowerCase().match( regex ); // eslint-disable-line
+	const [ , majorB, minorB, patchB, rcB ] = b.toLowerCase().match( regex ); // eslint-disable-line
```
tomalec: :grey_question: Why do we disable eslint here? And which one?
;) 
```suggestion
	const [ , majorA, minorA, patchA, rcA ] = a.toLowerCase().match( regex ); // eslint-disable-line -- reason why
	const [ , majorB, minorB, patchB, rcB ] = b.toLowerCase().match( regex ); // eslint-disable-line -- reason why
```

tomalec:
I get the point in counting the versions in a way `[7.7.rc-1, 7.6, 7.5, 7.4]` count to 4.
But I think a more usable case is to count it as 3. First of all, if we're L-2, we need to test 3 latest releases, so 7.6-7.4. Even when RC is published, we still need to support 7.4, (until the full `7.7` gets published).

Also, the counter is called `releases` and described as _"Number of **releases** to get."_ and `7.7.rc-1` is a release candidate, not a release :) So in my understanding, if we count "releases" not "versions"/"tags", the `releases: 3` setting should always include 3 full releases.

Then we can slightly adjust the wording in `includeRC` param and description to emphasize it's something added extra, like:
```yaml
  appendRC:
    description: Whether to add a potential Release Candidate to the list of fetched versions. (WordPress is Not compatible with this)
```

packages/js/github-actions/actions/get-plugin-releases/README.md
```diff
@@ -0,0 +1,64 @@
...
+
+#### Basic:
+
+```yaml
+on:
+  push:
+    branches:
+      - release/my-tool
+
+  jobs:
```
eason9487: The `jobs` probably don't need to be indented

packages/js/github-actions/actions/get-plugin-releases/README.md
```diff
@@ -0,0 +1,64 @@
...
+  GetPluginReleases:
+    name: Get Plugin Releases
+    runs-on: ubuntu-latest
+    steps:
+      - name: Get Release versions from WooCommerce
+        id: wc-matrix
+        uses: woocommerce/grow/get-plugin-releases@actions-v1.5.0-pre
```
eason9487: It would be proper to use the regular version as examples, such as `woocommerce/grow/get-plugin-releases@actions-v1`.

packages/js/github-actions/actions/get-plugin-releases/action.yml
```diff
@@ -0,0 +1,24 @@
...
+    default: "3"
+  includeRC:
+    description: Whether to include Release Candidates in the fetched versions. (WordPress is Not compatible with this)
+    default: "false"
+  includePatches:
+    description: Whether to include Patches in the fetched versions. (WordPress is NOT compatible with this)
+    default: "false"    
+    
+outputs:
+  matrix:
```
eason9487: Maybe change to using `versions` or some other name to avoid using `matrix` as the output name. Because `matrix` is also the name used in GitHub Actions to [define variations at a higher level for each job](https://docs.github.com/en/actions/using-jobs/using-a-matrix-for-your-jobs#about-matrix-strategies).

In addition to eliminating ambiguity, this action is named `get-plugin-releases`, so perhaps it could be expanded in the future to return more information related to plugin releases. By then, using `versions` would make it less confusing than `matrix` with the new naming.

packages/js/github-actions/actions/get-plugin-releases/src/get-plugin-releases.js
```diff
@@ -0,0 +1,141 @@
...
+				( includeRC || ! isRC( version ) ) &&
+				( includePatches || ! isMinorAlreadyAdded( output, version ) )
+			) {
+				output.push( version );
+			}
+		}
+	} else {
+		for ( const release of releases.offers ) {
+			if ( output.length === numberOfReleases ) {
+				return;
```
eason9487: This `return;` was missed to replace with `break;` in c18566fbd8cde0db6772597350e12f4782a42554.

image

See https://github.com/woocommerce/grow/actions/runs/4893825701/jobs/8737236994#step:8:13

packages/js/github-actions/actions/get-plugin-releases/src/get-plugin-releases.js
```diff
@@ -0,0 +1,141 @@
...
+	if ( ! rcA ) return -1;
+	if ( ! rcB ) return 1;
+
+	if ( rcB < rcA ) {
+		return -1;
+	}
+
+	if ( rcB > rcA ) {
+		return 1;
+	}
```
eason9487: Although in practice it hardly ever happens to have more than 9 RC versions with the same versioning number, in terms of version format, it's still not recommended to compare strings with the Greater than operator or Less than operator.

```js
'-rc.9' < '-rc.10'
// false

'-rc.10' > '-rc.9'
// false
```

packages/js/github-actions/actions/get-plugin-releases/README.md
```diff
@@ -0,0 +1,64 @@
+# Get latest releases versions from a specific plugin or from wordpress core
```
eason9487: This action currently only supports plugins that have been published to https://wordpress.org/. Other plugins, for example, the AutomateWoo plugin, are not supported.

It's recommended to describe the supported scope here, and perhaps also the `slug` parameter in action.yml.

eason9487 (commented):
Thanks for the work.

I did some local testing, however, the main issue is that getting versions with `wordpress` slug doesn't work. Recommend considering some adjustments for the rest comments.

puntope:
Hi @eason9487 @mikkamp @tomalec 

All the suggestions and bugs were fixed. I rebuilt the PRE release and [run a test workflow](https://github.com/woocommerce/grow/actions/runs/5013348553/jobs/8986318561).  

This is ready for another round.

packages/js/github-actions/actions/get-plugin-releases/README.md
```diff
@@ -0,0 +1,64 @@
...
+      - release/my-tool
+
+jobs:
+  GetPluginReleases:
+    name: Get Plugin Releases
+    runs-on: ubuntu-latest
+    steps:
+      - name: Get Release versions from WooCommerce
+        id: wc-versions
+        uses: woocommerce/grow/get-plugin-releases@actions-v1.5.0
```
eason9487: Setting the version number with `@actions-v1` would be better without the minor and patch parts. The [github-actions-release.yml](https://github.com/woocommerce/grow/blob/925104f9a80f183c68a1f0dfeb4634b0209a4987/.github/workflows/github-actions-release.yml#L131-L136) workflow will align `@actions-v1` with the latest version after release, so that once a grow GH action getting a minor or patch update,  the caller repos can continue to use the latest version without the needs to update the version tags.

packages/js/github-actions/actions/get-plugin-releases/src/get-plugin-releases.js
```diff
@@ -0,0 +1,140 @@
...
+	const [ , majorB, minorB, patchB, rcB ] = bMatches;
+
+	if ( majorA !== majorB ) return majorB - majorA;
+	if ( minorA !== minorB ) return minorB - minorA;
+	if ( patchA !== patchB ) return patchB - patchA;
+
+	if ( ! rcA ) return -1;
+	if ( ! rcB ) return 1;
+
+	return rcB.localeCompare( rcA );
```
eason9487: The concern in https://github.com/woocommerce/grow/pull/50#discussion_r1190960460 still exists.

For example, `[ '1.0.0-rc.1', '1.0.0-rc.9', '1.0.0-rc.10' ].sort( semverCompare )`
Expected: `['1.0.0-rc.10', '1.0.0-rc.9', '1.0.0-rc.1']`
Actual: `['1.0.0-rc.9', '1.0.0-rc.10', '1.0.0-rc.1']`

eason9487 (commented):
Thanks for the fixes and adjustments. However, two of the concerns mentioned in the previous code review still exist or could be further adjusted.

packages/js/github-actions/actions/get-plugin-releases/README.md
puntope: Understood, I didn't know about that version mapping feature. 

Adjusted here https://github.com/woocommerce/grow/pull/50/commits/73263f686e57e472142a9e5906ec96843e962f31

packages/js/github-actions/actions/get-plugin-releases/src/get-plugin-releases.js
puntope: Thanks again @eason9487  

I adjusted again the code here:

https://github.com/woocommerce/grow/pull/50/commits/1a2e777dba394246d1aff3975504680a50faf8b2

puntope:
Hello, @eason9487  Thanks for the previous suggestions. I patched them and this is ready for another round of reviews.

Thanks in advance

packages/js/github-actions/actions/get-plugin-releases/src/get-plugin-releases.js
```diff
@@ -0,0 +1,143 @@
...
+	if ( majorA !== majorB ) return majorB - majorA;
+	if ( minorA !== minorB ) return minorB - minorA;
+	if ( patchA !== patchB ) return patchB - patchA;
+
+	if ( ! rcA ) return -1;
+	if ( ! rcB ) return 1;
+
+	return (
+		parseInt( rcB.replace( '-rc.', '' ), 10 ) -
+		parseInt( rcA.replace( '-rc.', '' ), 10 )
```
eason9487, line 137: Maybe considering change the regex pattern to `/^(\d+)\.(\d+)\.(\d+)(?:-rc\.(\d+))?$/`, so that the two `.replace( '-rc.', '' )` here could be omitted.

eason9487 (approved):
Thanks for the work. LGTM.

Only a very minor thing might be considered to adjust, but it's okay to keep it as it is.

packages/js/github-actions/actions/get-plugin-releases/src/get-plugin-releases.js
puntope: Since its very minor and now its working and finally approved I will keep it for now like this.

PR #52, opened 2023-04-20
Grow Smooth Generator

mikkamp:

This was a hack day project to create a plugin which can generate data for extensions that are maintained by Grow. The plugin is based off of the [WooCommerce Smooth Generator](https://github.com/woocommerce/wc-smooth-generator#readme) which must be activated on the site.

Initially I focused on setting up a basic scaffold for the plugin and then started with adding commands for generating AutomateWoo data. To limit the scope of this project I only implemented WP CLI commands. I was able to add the following two generators for AutomateWoo:

- Conversion data
- Email and SMS tracking data (opens and clicks)

I didn't get round to including any build process or packaging the plugins into a zip file. So for now the install must be done manually.

### Example output:
```
$ wp wc generate aw-email-tracking 10 --workflow=142 --date-start=2023-04-01 --date-end=2023-04-10
Initializing...
Generating email and SMS tracking  100% [================================================================================] 0:00 / 0:00
Success: 10 email and SMS tracking generated in 0.05 seconds
```

1. As prerequisites make sure both the [WooCommerce Smooth Generator](https://github.com/woocommerce/wc-smooth-generator) plugin and AutomateWoo are installed and activated on the site
2. Clone the grow repo locally 
3. Create a symbolic link in your WordPress plugins directory `wp-content/plugins` to the folder `plugins/grow-smooth-generator`
4. In the plugin folder run `composer install`
5. Activate the plugin `wp plugin activate grow-smooth-generator`
6. Generate a workflow in AutomateWoo and make sure to check "Enable tracking" and "Enable conversion tracking"
image
7. Run a WP CLI command to generate some conversion data `wp wc generate aw-conversions 10 --workflow=<workflow ID> --date-start=2023-04-01 --date-end=2023-04-10`
8. Go to Analytics > Conversions and select a report for the month of April, confirm that we can see the generated conversion data from the 1st till the 10th of April:
image
9. Run a WP CLI command to generate email tracking data `wp wc generate aw-email-tracking 10 --workflow=<workflow ID> --date-start=2023-04-01 --date-end=2023-04-10`
10. Go to Analytics > Email & SMS Tracking and select a report for the month of April, confirm that we can see the generated opens/clicks from the 1st till the 10th of April:
image

> **Note**
If you viewed the report for the month of April (causing the data to be cached), and then generated the new data, then it might not show up right away in the results. This is because WC Analytics doesn't have a way to clear the cache for a set of reports, we have an open issue for this here: https://github.com/woocommerce/automatewoo/issues/1355
To get around this you can clear the cache manually or select a different date range in the reports.

### Additional Information
The plugin has been structured to easily add additional commands for AutomateWoo as well as add generators for other extensions.

See `includes/CLI.php` for adding an additional extension, an additional class must also be created in `includes/CLI` for registering the extensions commands.
Individual generators can be created in `includes/Generator` and can be group in a sub folder for each extension.

### Changelog entry
* Add - Create a Grow Smooth Generator plugin.


plugins/grow-smooth-generator/includes/CLI.php
```diff
@@ -0,0 +1,19 @@
...
+ */
+class CLI extends WP_CLI_Command {
+
+	/**
+	 * Conditionally load commands for each extension.
+	 */
+	public function __construct() {
+		if ( defined( 'AUTOMATEWOO_VERSION' ) ) {
+			CLI\AutomateWoo::add_commands();
+		}
```
puntope, line 17: 💅  Minor, but I would add this inside `CLI\AutomateWoo::add_commands(); ` so CLI class doesn't need to worry about each command requirements.

plugins/grow-smooth-generator/includes/CLI/AutomateWoo.php
```diff
@@ -0,0 +1,150 @@
...
+			$progress->tick();
+		}
+
+		$time_end       = microtime( true );
+		$execution_time = round( ( $time_end - $time_start ), 2 );
+		$display_time   = $execution_time < 60 ? $execution_time . ' seconds' : human_time_diff( $time_start, $time_end );
+
+		$progress->finish();
+
+		WP_CLI::success( $amount . ' email and SMS tracking generated in ' . $display_time );
```
puntope, line 72: This code is almost the same as `AutomateWoo::conversions`, and each time we add a command is going to be likely duplicated again. 

Maybe we can wrap this over something like `AutomateWoo::run( $generator, $args, $assoc_args )`

plugins/grow-smooth-generator/includes/Generator/AutomateWoo/EmailTracking.php
```diff
@@ -0,0 +1,136 @@
...
+		}
+
+		$dates = array();
+		$date  = strtotime( $start );
+		while ( $date <= strtotime( $end ) ) {
+			$dates[] = date( 'Y-m-d', $date );
+			$date    = strtotime( '+1 day', $date );
+		}
+
+		$date = $dates[ array_rand( $dates ) ];
```
puntope: Not sure but I think this is to generate a random date between the interval.??? Another alternative without the while might be:

```
    $date_start = strtotime($start);
    $date_end = strtotime($end);
    $date = rand($date_start, $date_end);

    $date =  date('Y-m-d', $date);
    
    if ( $include_time ) {
			return $date . ' ' . rand( 0, 23 ) . ':00:00';
		}

		return $date;
```

plugins/grow-smooth-generator/README.md
```diff
@@ -0,0 +1,35 @@
...
+2. From command line CD into the plugin folder
+3. Run `composer install` and wait for the installation to complete
+4. Run `wp plugin activate grow-smooth-generator` to activate the plugin
+5. You now have access to a couple of new WP-CLI commands under the main `wp wc generate` command.
+
+## Commands
+
+### AutomateWoo Conversions
+
+Generate conversions based on the number of conversions parameter.
```
puntope: 💅  Maybe is worth it to clarify that without the Date params, we are loading the data for the current date.

puntope (commented):
Hello @mikkamp 

Thanks for working on this, it will be useful for testing our extensions.

The behaviour works as expected, I left some comments regarding the code tho. 

Since you mentioned is planned to extend the code with more generators I think we should at least refactor the "progress" code block inside `conversions` and  `email_tracking` to avoid duplication. 

The while loop + random array strategy to get a random date also was a bit confusing for me. But that's subjective I guess 💅 .

plugins/grow-smooth-generator/includes/CLI/AutomateWoo.php
mikkamp: While I do agree there are quite a few similarities, there are still enough substantial differences to make an additional wrapper cumbersome. For example in one it halts order emails, also in this case they both accept a date range, but that's not the case with every command added in the future.

So in this case I chose to go with some duplication to make it easy to see exactly which actions are run per command. It's also the structure which is used for the WC Smooth Generator.

plugins/grow-smooth-generator/includes/Generator/AutomateWoo/EmailTracking.php
mikkamp: Yes that makes sense, I copied this code 1:1 from WC Smooth Generator, so I didn't put much thought into it. I'll go ahead and add your suggestion though.

plugins/grow-smooth-generator/includes/CLI.php
mikkamp: Pretty much all our extensions have a defined version so it's really only one requirement check. The only thought I had against your point is that means all the classes will have to be autoloaded to check the conditions. Whereas the way it is now it will only autoload the AutomateWoo class if the plugin is active.

mikkamp:
Thanks for the review, I've addressed those comments and made some tweaks. Let me know if that works, it's ready for another round of review.

plugins/grow-smooth-generator/includes/Generator/AutomateWoo/EmailTracking.php
puntope: Since now we included this way to calculate the dates we can remove the old code:

```
$dates = array();
		$date  = strtotime( $start );
		while ( $date <= strtotime( $end ) ) {
			$dates[] = date( 'Y-m-d', $date );
			$date    = strtotime( '+1 day', $date );
		}

		$date = $dates[ array_rand( $dates ) ];
```

puntope (commented):
Thanks @mikkamp  for the changes.

Seems like the old code after the changes in https://github.com/woocommerce/grow/pull/52/commits/f5df9aad531c7fc54c30b0be571a1bcd2fe7dec5 is still there and should be removed.

mikkamp:
> Seems like the old code after the changes in https://github.com/woocommerce/grow/commit/f5df9aad531c7fc54c30b0be571a1bcd2fe7dec5 is still there and should be removed.

Ha, good observation, guess I forgot something there. Fixed in aeee9e5

puntope (approved):
✅  LGTM 

Thanks @mikkamp

PR #53, opened 2023-05-12
Fix the missed `jobs` and incorrect `uses` in the README of `get-release-notes` action

eason9487:

While reviewing #50, found a few mistakes in the `get-release-notes` action document. This PR fixes:
- The missed `jobs` key in an example.
- Fix the incorrect action URIs and tags of the `uses` in all examples.

View the [get-release-notes/README](https://github.com/woocommerce/grow/blob/aebc841df86be499abd64ce3a764a9177f409561/packages/js/github-actions/actions/get-release-notes/README.md) to check for corrections.


tomalec (approved):
LGTM thanks! :)

PR #54, opened 2023-05-19
Tweak - Upgrade actions/core to v1.10.0

puntope:

[Set Output command is being deprecated](https://github.blog/changelog/2022-10-11-github-actions-deprecating-save-state-and-set-output-commands/) soon and is producing a warning in our JS based actions using SetOutput

For solving it, this PR updates actions/core to v1.10.0 as suggested in the GitHub blog.

Besides that, updating the actions/core package produces an error with rollup build file that is solved by adding `exportConditions: [ 'node' ]` inside the `nodeResolve` plugin to correctly use the node version.

1. Confirm that get-plugin-releases action works fine
2. Confirm that get-release-notes action works fine
3.  Confirm that update-version-tags action works fine
4. Confirm that php-diff action works fine
5. Confirm no warning is displayed anymore  (in  update-version-tags  and php-diff) this only happens when errors happen. Since core.setFailed  is called here only)

ℹ️  The release process was tested here: https://github.com/puntope/grow/actions/runs/5055625218/jobs/9072072131
ℹ️  See also https://github.com/puntope/grow/pull/1

### Changelog entry

> Tweak - Upgrade actions/core to v1.10.0


eason9487 (approved):
Thanks for the update. LGTM.

The release process will only use PR titles as the changelog entities, so the following note won't be used. Maybe describe the change note in this PR title instead.

> ### Changelog entry
> > Tweak - Upgrade actions/core to v1.10.0

eason9487:
After I ran `npm install` with node v16, the package-lock.json file will have a few more changes. Maybe these changes need to be added as well?

puntope:
> After I ran `npm install` with node v16, the package-lock.json file will have a few more changes. Maybe these changes need to be added as well?

Supposedly it should be run by Node14 since it's the defined version in the `.nvmrc`

NEw package-lock.json was added here https://github.com/woocommerce/grow/pull/54/commits/212d53c82c0027b450f3d74f1fffffcc13802c2f

PR #58, opened 2023-06-07
Add pre- & post-steps to extension release PR

tomalec:

- [Rename prepare-extension-release inputs to kebabcase,](https://github.com/woocommerce/grow/pull/58/commits/0cf5cb5d97395e76b7131a4df0c7a02456a745c4) 
   to follow the convention of other actions.
   Read them as JSON.
- [Add pre- & post-steps inputs,](https://github.com/woocommerce/grow/pull/58/commits/35882f69ff6e5d7bd075a2607dd1d6b514e45e53) 
   to be able to customize a little the release proces per each consuming repo.
- [Return outputs, generate a summary](https://github.com/woocommerce/grow/pull/58/commits/1d06aca71023fceac8f0a6c40bac81f1d3bde6fe) 
   with a link to the created PR.

### Screenshots:
image

1. Make a workflow that uses `woocommerce/grow/prepare-extension-release@actions-v1.6.1-pre`, its results, and  `pre-`, `post-steps`,
   like https://github.com/tomalec/grow/blob/a1a62cfb29783d9636576bfb354a557254529414/.github/workflows/extension-release.yml#L36-L70
3. [Run that workflow](https://github.com/tomalec/grow/actions/runs/5214987470)
4. Check that workflow has a PR link in the summary
5. Check that the resulting PR has additional steps https://github.com/tomalec/grow/pull/33

### Changelog entry

> Add `pre-` & `post-steps` to `prepare-extension-release` action, return some outputs.

### DIFF
--- a/packages/js/github-actions/actions/prepare-extension-release/README.md
+++ b/packages/js/github-actions/actions/prepare-extension-release/README.md
@@ -30,8 +30,9 @@ on:
       wc-version:
         description: 'WooCommerce tested up to'
 
+jobs:
   Prepare_release:
-    name: Prepare release
+    name: 'Prepare release'
     runs-on: ubuntu-latest
     steps:
       - name: Checkout repository
--- a/packages/js/github-actions/actions/prepare-extension-release/README.md
+++ b/packages/js/github-actions/actions/prepare-extension-release/README.md
@@ -25,9 +25,9 @@ on:
         description: 'Type of the release (release|hotfix)'
         required: true
         default: 'release'
-      wp_version:
+      wp-version:
         description: 'WordPress tested up to'
-      wc_version:
+      wc-version:
         description: 'WooCommerce tested up to'
 
   Prepare_release:
@@ -40,7 +40,7 @@ on:
         with:
           version: ${{ github.event.inputs.version }}
           type: ${{ github.event.inputs.type }}
-          wp_version: ${{ github.event.inputs.wp_version }}
-          wc_version: ${{ github.event.inputs.wc_version }}
-          main_branch: 'trunk'
+          wp-version: ${{ github.event.inputs.wp-version }}
+          wc-version: ${{ github.event.inputs.wc-version }}
+          main-branch: 'trunk'
 ```
--- a/packages/js/github-actions/actions/prepare-extension-release/README.md
+++ b/packages/js/github-actions/actions/prepare-extension-release/README.md
@@ -43,4 +43,10 @@ on:
           wp-version: ${{ github.event.inputs.wp-version }}
           wc-version: ${{ github.event.inputs.wc-version }}
           main-branch: 'trunk'
+          pre-steps: |
+            1. [ ] Prepare something more before the release
+          post-steps: |
+            ### Additional post-release checklist
+            1. [ ] Update documentation
+               - [ ] Publish any new required docs
 ```
--- a/packages/js/github-actions/actions/prepare-extension-release/src/woo-extension-create-pr-for-release.mjs
+++ b/packages/js/github-actions/actions/prepare-extension-release/src/woo-extension-create-pr-for-release.mjs
@@ -1,13 +1,11 @@
-export default async ( {
-	github,
-	context,
-	base,
-	refName,
-	type,
-	version,
-	wpVersion,
-	wcVersion,
-} ) => {
+export default async ( { context, github, inputs, refName } ) => {
+	const {
+		'main-branch': base,
+		type,
+		version,
+		'wc-version': wcVersion,
+		'wp-version': wpVersion,
+	} = inputs;
 	// Assume the extension package is named as the repo.
 	const extensionPackageName = context.payload.repository.name;
 	// Build repo URL. WooRelease expects HTML one with `https://github.com/…/tree…`.
--- a/packages/js/github-actions/actions/prepare-extension-release/src/woo-extension-create-pr-for-release.mjs
+++ b/packages/js/github-actions/actions/prepare-extension-release/src/woo-extension-create-pr-for-release.mjs
@@ -1,6 +1,8 @@
 export default async ( { context, github, inputs, refName } ) => {
 	const {
 		'main-branch': base,
+		'post-steps': postSteps,
+		'pre-steps': preSteps,
 		type,
 		version,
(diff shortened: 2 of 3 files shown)


packages/js/github-actions/actions/prepare-extension-release/README.md
```diff
@@ -40,7 +40,13 @@ on:
         with:
           version: ${{ github.event.inputs.version }}
           type: ${{ github.event.inputs.type }}
-          wp_version: ${{ github.event.inputs.wp_version }}
-          wc_version: ${{ github.event.inputs.wc_version }}
-          main_branch: 'trunk'
+          wp-version: ${{ github.event.inputs.wp-version }}
+          wc-version: ${{ github.event.inputs.wc-version }}
```
mikkamp, line 45: I can see this matches the inputs in the original workflow:

```
          wp-version: ${{ github.event.inputs.wp-version }}
          wc-version: ${{ github.event.inputs.wc-version }}
```

But in the final simulate / release commands the versions are missing:
```
woorelease simulate --product_version=1.2.3  --generate_changelog https://github.com/mikkamp/tryout/tree/release/1.2.3
```

Is a dash in a variable name allowed? From a quick search I'm seeing the following might be needed:
```
${{ github.event.inputs['wp-version'] }}
```

packages/js/github-actions/actions/prepare-extension-release/src/woo-extension-create-pr-for-release.mjs
```diff
@@ -19,9 +19,15 @@ export default async ( {
...
+		trimmedPreSteps = '\n' + trimmedPreSteps.trim();
+	}
+
 	const body = `## Checklist
 1. [ ] Check if the version, base, and target branches are as you desire.
-1. [ ] Make sure you have \`woorelease\` installed and set up.
+1. [ ] Make sure you have \`woorelease\` installed and set up.${ trimmedPreSteps }
 1. [ ] Simulate the release locally
    \`\`\`sh
    git fetch origin ${ refName }
```
mikkamp, line 33: Just curious why these `git fetch` and `git checkout` lines are added here. Those are needed if we want to update our local copy, but not needed for woorelease since it does a separate clone/checkout to the tmp folder. Not bad to leave but just wonder if it needs additional clarification as those commands need to be run specifically in a local repo copy.

mikkamp (commented):
Just a left a couple comments as I went through testing this. But unable to test fully as mentioned in the other PR.

packages/js/github-actions/actions/prepare-extension-release/README.md
tomalec: I tried it in https://github.com/tomalec/grow/blob/b1448e10c86b1e4460bf2e9974911ab89f9cd692/.github/workflows/extension-release.yml#L19-L41
and it seems to work correctly

https://github.com/tomalec/grow/pull/35

could you make your repo public?

tomalec: Maybe it's better to change to camelCase, so it will work for sure and will not require mapping in `js` files.

Or we can change to the snake_case, to follow GitHub workflow convention (which is still a bit inconsistent and conducting to me as it [has `branches-ignore`](https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#example-excluding-branches))

packages/js/github-actions/actions/prepare-extension-release/src/woo-extension-create-pr-for-release.mjs
tomalec: Good point. I kept them here, as some repos require additional steps to perform on the repo like GLA requires manual removal of entries from `changelog.txt`. 
I wonder what's better:
- remove it so other repos will have a cleaner process
- keep it, it's copied together with the simulate anyway and does no harm
- automate the removal of changelog.txt :godmode: & co.

packages/js/github-actions/actions/prepare-extension-release/README.md
tomalec: Are you sure you used 
```yml
uses: woocommerce/grow/prepare-extension-release@actions-v1.6.1-pre
```
not `v1`?

I tested in on a fresh repo and it works fine https://github.com/tomalec/test-dashes-in-workflows/actions/runs/5265434320

packages/js/github-actions/actions/prepare-extension-release/src/woo-extension-create-pr-for-release.mjs
tomalec: > remove it so other repos will have a cleaner process

If we remove it from here, in additional steps we will not have the possibility to provide the exact `refName` (unless we introduce a templating engine, which IMO is overkill, given our goal is to simplify and unify steps to be same for all repos)

tomalec: Addressed somewhat in https://github.com/woocommerce/grow/pull/58/commits/6256683d661e05461ef7fa2a8023b8915591181d

tomalec:
Thanks @mikkamp, for the review.

I was not able to reproduce the problem with dashes. Would you mind taking another look?

packages/js/github-actions/actions/prepare-extension-release/README.md
mikkamp: It would help if I read the full testing instructions. I was originally going by the steps here: https://github.com/woocommerce/google-listings-and-ads/pull/1982

And assumed you were going to update the branch there when the PR got merged, so I didn't double check. Switching it out to the branch allows everything to work. I'll check the PR's again.

mikkamp (approved):
Thanks, for the clarifications on how to run this. It's working as expected now, it didn't include your last commit since it's not included in the pre-release, but the change looks good.

packages/js/github-actions/actions/prepare-extension-release/README.md
tomalec: Ah, Sorry I didn't mention it there :)

tomalec:
After https://github.com/woocommerce/automatewoo/pull/1451/files is merged we can release this PR as a new version

PR #66, opened 2023-07-21
Get Plugin releases Action: Fix error counting RC releases

### LINKED ISSUE #65: Fix condition when loading RC versions.
Ref: p1689857980355099/1689849074.370359-slack-C0410KV3YLW

It seems we need to tweak the condition at
[https://github.com/woocommerce/grow/blob/e0ee37b89343ba9b6b4d6db6088b3a3a4ecefe50/[…]-actions/actions/get-plugin-releases/src/get-plugin-releases.js](https://github.com/woocommerce/grow/blob/e0ee37b89343ba9b6b4d6db6088b3a3a4ecefe50/packages/js/github-actions/actions/get-plugin-releases/src/get-plugin-releases.js#L67)
to the same we have for pushing to outputs: 

`release => (includeRC || !isRC(release)),`

Then latest-versions `woocommerce 5 --includePatches --includeRC` would return `["7.9.0","7.9.0-rc.3","7.9.0-rc.2","7.8.2","7.8.1","7.8.0"]`

puntope:

Closes #65 

RC was not being counted as the output for plugins other than WordPress. So when we added Including RC option they were adding more releases than expected. 

1.  Check if the [Pre Release is Compiling](https://github.com/woocommerce/grow/actions/runs/5621686842) 
2.  Check if the [action outputs make sense](https://github.com/woocommerce/grow/actions/runs/5621821892). 

### Changelog entry

> Fix - Error counting RC releases.


eason9487 (approved):
LGTM.

Considering the PR title will be used as an entry of release note, it would be a bit cleaner if the title also indicates which action to be fixed.

PR #79, opened 2023-08-28
Add Compat Checker library for versioning and compatibility checks

### LINKED ISSUE #75: Implement Compat Checker PHP package
## ~User~Developer story

As a Grow developer, I'd like to run plugin the following compatibility checks across all our repos from a single composer package:

* WooCommerce minimum compatibility
* WooCommerce L-N support notice
* WordPress minimum compatibility
* PHP compatibility (planned)

### Is your feature request related to a problem? 

To enforce the L-N support policy, we need to run checks before initialising the plugin and run helpful notices or stop the execution of the plugin.

### Describe the solution you'd like 

Build a composer package with a `CompatChecker` class that runs all the compatibility checks based on the plugin header file. The compat-checker would be similar to the [`PluginValidator`](https://github.com/woocommerce/google-listings-and-ads/blob/develop/src/Internal/Requirements/PluginValidator.php) implemented in Google Listings and Ads.

The composer package can reside in a separate branch, and then be included in the individual plugins. We can use a GitHub action like this: https://github.com/marketplace/actions/push-git-subdirectory-as-branch to automatically deploy the changes into the `packages/compat-checker` folder into the separate branch.

### Describe alternatives you've considered 

The alternative would be individually implementing a `PluginValidator` class across all our plugins. This would be redundant.

ibndawood:

Closes #75.

This PR:

* Adds a new `PHP` package called `compat-checker` that can be installed via composer in WooCommerce Grow extensions.
* This new package runs compatibility checks based on the plugin's header file.
* Here is a list of compatibility checks run:
    * Checks if the WooCommerce plugin is installed and active.
    * Checks if the active WooCommerce version meets the minimum WooCommerce version defined in the plugin header.
    * Displays a warning if the active WooCommerce version is behind L-N (2 is the default) version.
    * Displays a warning if the active WooCommerce is untested with the plugin.
    * Displays a warning if the current WordPress version is untested with the plugin.

#### Setting up the fork.

1. Open the WooCommerce Grow repository from the URL: https://github.com/woocommerce/grow
2. On the top right, click the **Fork** button to fork your own copy.
3. Ensure the `Copy the trunk branch only` is unchecked.
4. Click the `Create Fork` button. This step will create a fork of [`woocommerce/grow`](https://github.com/woocommerce/grow). Let's say: [`ibndawood/grow`](https://github.com/ibndawood/grow/)
5. In the forked repository, create a new PR to merge the branch `add/compat-checker` branch into `trunk`.
6. Merge the PR.

#### Testing the GitHub Workflow

1. Navigate to `Actions` tab in the forked repository. 
2. Ensure `Public Compat Checker package` workflow is added.
3. Ensure there is a branch named `compat-checker`.
4. Make a simple commit to the `trunk` to trigger the workflow.
5. Ensure the subdirectory `packages/php/compat-checker` contents are published into the `compat-checker` branch.

#### Including the library in the WooCommerce Extension

1. Clone any one of the WooCommerce Grow extensions. Let's say: [WooCommerce Brands](https://github.com/woocommerce/woocommerce-brands/)
2. Include the `compat-checker` package by adding the repository to the `composer.json` of the cloned repository like in the snippet below. Replace `https://github.com/ibndawood/grow` with the URL of your fork:
```json
"repositories": [
    {
        "type": "vcs",
        "url": "https://github.com/ibndawood/grow"
    }
],
"require": {
    "woocommerce/grow": "dev-compat-checker"
},
```
3. Run `composer update` to included `woocommerce/grow` in the plugin's `vendor` folder.
4. Ensure the `vendor` folder has the `woocommerce/grow` folder.
5. Ensure the `vendor/composer/autoload_psr4.php` file has `Automattic\WooCommerce\Grow\Tools\` namespace included.

##### Setup
In the main plugin file, run `Automattic\WooCommerce\Grow\Tools\CompatChecker\v0_0_1\Checker::instance()->is_compatible( __FILE__, $plugin_version )` to check for compatibility and versioning. The snippet should be placed before initialising the plugin. The below example snippet is for WooCommerce Brands:
```php
require __DIR__ . '/vendor/autoload.php';
use Automattic\WooCommerce\Grow\Tools\CompatChecker\v0_0_1\Checker;

// Plugin init hook.
add_action( 'plugins_loaded', 'wc_brands_init', 1 );

/**
 * Initialize plugin.
 */
function wc_brands_init() {
    define( 'WC_BRANDS_VERSION', '1.6.56' ); // WRCS: DEFINED_VERSION.

    if ( ! Checker::instance()->is_compatible( __FILE__, WC_BRANDS_VERSION ) ) {
	return;
    }
    // Continue with initializing...
}
```
##### WooCommerce activation tests

With the above code snippets in place, run the following tests. The instructions are for WooCommerce Brands but can be used for other WooGrow Extensions.

1. Install and activate WooCommerce Brands without installing WooCommerce.
2. Deactivate WooCommerce.
3. Verify the notice `WooCommerce Brands requires WooCommerce to be installed and activated. Please install WooCommerce.` is displayed.
4. Install WooCommerce without activating.
5. Verify the notice `WooCommerce Brands requires WooCommerce to be activated. Please activate WooCommerce.` is displayed.

##### WooCommerce versioning tests
⚠️  **Important note**: You will need to erase the transient `_transient_wc_grow_compat_checker_[EXTENSION].php[VERSION` after each change!!

1. Install and activate WooCommerce.
2. Install and activate WooCommerce Brands.
3. Modify the `WC requires at least` header field and set it to a version higher than the active WooCommerce version. Let's say the `WC requires at least` is set to `8.2` and the active WooCommerce version is `8.0.2`.
4. Verify the notice `WooCommerce Brands requires WooCommerce version 8.2 or higher. Please update WooCommerce to the latest version, or download the minimum required version` is displayed.
5. Modify the `WC tested up to` and set it to a lower version than the active WooCommerce version. Let's say the `WC tested up to` is set to `7.8`, the active WooCommerce version is `8.0.2` and WooCommerce Brands version is `1.6.56`.
6. Verify the warning notice `WooCommerce Brands - 1.6.56 is untested with WooCommerce 8.0.2` is displayed.

##### L2 support recommendation notice

1. Download the latest minus three WooCommerce version and activate it. Let's say `7.6`.
2. Verify the warning notice `Heads up! WooCommerce Brands will soon discontinue support for WooCommerce 7.6. Please update WooCommerce to take advantage of the latest updates and features.`

##### WordPress versioning tests
⚠️  **Important note**: You will need to erase the transient `_transient_wc_grow_compat_checker_[EXTENSION].php[VERSION` after each change!!

1. Modify the `Tested up to` and set it to a lower version than the current WordPress version. Let's say the `Tested up to` is set to `6.2`, the active WordPress version is `6.3`, and the WooCommerce Brands version is `1.6.56`.
2. Verify the warning notice `WooCommerce Brands - 1.6.56 is untested with WordPress 6.3` is displayed.

### Changelog entry

> Add - Compat Checker library for versioning and compatibility checks.


mikkamp:
This looks like an awesome project and way to simplify code across extensions.

We used to have a PHP utilities file which was added to every plugin with compatibility checks similar but not quite as elaborate as these. But since it wasn't a composer package it was hard to maintain in a central location, so it got removed.

I haven't gone through all the code, but one thing I noticed is the difference with how currently a lot of checks are done by comparing several consts to ensure the version is right (which would be relatively low impact). However these checks include a call to `get_file_data`, and a (cached) call to `wordpress.org`. Since these checks would be done on every single request, is there some ways we can bring the resource usage down as low as possible? Like caching the `get_file_data` values (per plugin). I haven't tested, but from my impression reading a file (first 8Kb) and then parsing the header, would be slower than fetching a transient.
Also as far as I understand `check_wc_upgrade_recommendation` would be useful for the admin notices, but not on any frontend requests. Can that be moved to only backend, so the call to `wordpress.org` would never get triggered from a frontend request?

Another thing I was wondering is how this composer package will maintain versions.

For example:
```
plugin abeautifulworld loads compat-checker 0.0.1
plugin xmessaging loads compat-checker 0.0.2
```

Both plugins will now be using the code from 0.0.1 because plugins are loaded alphabetically. We are probably going to get minimal changes in the compat-checker code, so it might not be a big deal. But is there any concern about always loading the latest version (without resorting to a solution which might increase resource usage)?

budzanowski:
@mikkamp 

> I haven't tested, but from my impression reading a file (first 8Kb) and then parsing the header, would be slower than fetching a transient.

That is a correct observation. We have already discussed this with @ibndawood and I think that this is the next step.

> Both plugins will now be using the code from 0.0.1 because plugins are loaded alphabetically. We are probably going to get minimal changes in the compat-checker code, so it might not be a big deal. But is there any concern about always loading the latest version (without resorting to a solution which might increase resource usage)?

Maintaining API compatibility between packages versions will be critical.

ibndawood:
Thank you, @mikkamp, for taking the time to look into the PR and for the questions. 

> However these checks include a call to get_file_data, and a (cached) call to wordpress.org.

This is right. @budzanowski also raised this question during our 1:1s. I had incorrectly assumed that `get_file_data` was cached by WordPress internally. I was hoping to have discussions around this during the code review.

The call to `wordpress.org` made from `check_wc_upgrade_recommendation` should run in the backend only. I had overlooked this. Thank you for catching this. I'll correct this and shall push an update.

As for the `get_file_data`, I looked for examples and noticed that WooCommerce Blocks uses this approach to check for dependencies before the `woocommerce_blocks_loaded` action. They didn't use transients but we can use transients for subsequent requests. However, I wanted to discuss what would be the best expiration time for these transients. Your thoughts?

ibndawood:
> Another thing I was wondering is how this composer package will maintain versions.

Since we are installing from a branch, we can only specify the branch which is `dev-compat-checker` for versioning.

mikkamp:
> However, I wanted to discuss what would be the best expiration time for these transients. Your thoughts?

Every plugin we manage has a const version we manage with the comment `// WRCS: DEFINED_VERSION.`
Pass that const together with the file name and it can be used to generate a unique transient name, that will expire when there are changes to the file. Which means in theory it never needs to expire, but you could probably add some kind of "long expiry" time in case there is a cleanup routine to clear up old transients.

ibndawood:
> Every plugin we manage has a const version we manage with the comment // WRCS: DEFINED_VERSION.
Pass that const together with the file name and it can be used to generate a unique transient name, that will expire when there are changes to the file.

Thank you, this is a great idea. I'll refactor as suggested.

ibndawood:
I've added caching for `get_file_data` and limited `check_wc_upgrade_recommendation` to backend only.

ibndawood:
@mikkamp, @budzanowski - Resolved versioning issues by adding the version information to the namespace.

packages/php/compat-checker/src/Checker.php
```diff
@@ -0,0 +1,87 @@
...
+			'TestedWP'    => 'Tested up to',
+			'TestedWC'    => 'WC tested up to',
+		);
+
+		$transient_key = 'wc_grow_compat_checker_' . $plugin_file . $file_version;
+		$plugin_data   = get_transient( $transient_key );
+
+		if ( false === $plugin_data ) {
+			$plugin_data = get_file_data( $plugin_file, $default_headers, 'plugin' );
+			set_transient( $transient_key, $plugin_data, MONTH_IN_SECONDS );
```
layoutd, line 58: Is a month too much? What happens if I auto-upgrade an extension and the new version now requires WC 8.2, but I have WC 7.9? It won't catch that until the transient expires, and could throw errors.
Should these transients be erased for each extension on upgrade?

packages/php/compat-checker/src/Checks/WCCompatibility.php
```diff
@@ -0,0 +1,389 @@
...
+
+		$update_url = wp_nonce_url( self_admin_url( 'update.php?action=upgrade-plugin&plugin=' . self::WC_PLUGIN_FILE ), 'upgrade-plugin_' . self::WC_PLUGIN_FILE );
+		$message    = sprintf(
+			/* translators: %1$s - Plugin Name, %2$s - minimum WooCommerce version, %3$s - update WooCommerce link open, %4$s - update WooCommerce link close, %5$s - download minimum WooCommerce link open, %6$s - download minimum WooCommerce link close. */
+			esc_html__( '%1$s requires WooCommerce version %2$s or higher. Please %3$supdate WooCommerce%4$s to the latest version, or %5$sdownload the minimum required version &raquo;%6$s', 'woogrow-compat-checker' ),
+			'<strong>' . $plugin_name . '</strong>',
+			$wc_version_required,
+			'<a href="' . esc_url( $update_url ) . '">',
+			'</a>',
+			'<a href="' . esc_url( 'https://downloads.wordpress.org/plugin/woocommerce.' . $wc_version_required . '.zip' ) . '">',
```
layoutd, line 313: `$wc_version_required` could be a major-minor value (e.g., `8.1`), but the download links all have major-minor-patch nomenclature (e.g., `8.1.0`).

For example, I put ` * WC requires at least: 8.1`.
The generated link is <https://downloads.wordpress.org/plugin/woocommerce.8.1.zip> 🔴 
The correct link is <https://downloads.wordpress.org/plugin/woocommerce.8.1.0.zip> 🟢 

The version number should be padded with `.0` until it matches `\d+\.\d+\.\d+` I guess

packages/php/compat-checker/src/Checks/WCCompatibility.php
```diff
@@ -0,0 +1,389 @@
...
+		$plugin_name         = $this->plugin_data['Name'];
+		$wc_version_required = $this->plugin_data['RequiresWC'];
+
+		$update_url = wp_nonce_url( self_admin_url( 'update.php?action=upgrade-plugin&plugin=' . self::WC_PLUGIN_FILE ), 'upgrade-plugin_' . self::WC_PLUGIN_FILE );
+		$message    = sprintf(
+			/* translators: %1$s - Plugin Name, %2$s - minimum WooCommerce version, %3$s - update WooCommerce link open, %4$s - update WooCommerce link close, %5$s - download minimum WooCommerce link open, %6$s - download minimum WooCommerce link close. */
+			esc_html__( '%1$s requires WooCommerce version %2$s or higher. Please %3$supdate WooCommerce%4$s to the latest version, or %5$sdownload the minimum required version &raquo;%6$s', 'woogrow-compat-checker' ),
+			'<strong>' . $plugin_name . '</strong>',
+			$wc_version_required,
+			'<a href="' . esc_url( $update_url ) . '">',
```
layoutd: I think this is super cool, but I worry a bit about having the link directly update the plugin. What if I have made modifications to my WC install, or have a repo? I may click that link thinking it will take me to the download page or scroll me to the WooCommerce row in the plugins table, and all of a sudden my WooCommerce changes are gone without a confirmation or anything.

packages/php/compat-checker/src/Checker.php
```diff
@@ -0,0 +1,87 @@
...
+			'Name'        => 'Plugin Name',
+			'Version'     => 'Version',
+			'RequiresWP'  => 'Requires at least',
+			'RequiresPHP' => 'Requires PHP',
+			'RequiresWC'  => 'WC requires at least',
+			'TestedWP'    => 'Tested up to',
+			'TestedWC'    => 'WC tested up to',
+		);
+
+		$transient_key = 'wc_grow_compat_checker_' . $plugin_file . $file_version;
```
layoutd: 💅🏻  While very specific, the `$transient_key` value `_transient_timeout_wc_grow_compat_checker_/var/www/html/wp-content/plugins/woocommerce-brands/woocommerce-brands.php1.6.58` is not particularly attractive. Should this use `plugin_basename()` maybe?
E.g., `_transient_timeout_wc_grow_compat_checker_woocommerce-brands/woocommerce-brands.php1.6.58`

packages/php/compat-checker/src/Checks/WCCompatibility.php
```diff
@@ -0,0 +1,389 @@
...
+	 * Adds notice for untested WooCommerce.
+	 */
+	public function wc_untested() {
+		$plugin_name    = $this->plugin_data['Name'];
+		$wc_version     = $this->get_wc_version();
+		$plugin_version = $this->plugin_data['Version'];
+
+		$message = sprintf(
+			/* translators: %1$s - Plugin Name, %2$s - Plugin version, %3$s - WooCommerce version number */
+			esc_html__( '%1$s - %2$s is untested with WooCommerce %3$s.', 'woogrow-compat-checker' ),
```
layoutd: ❓ Should the merchant be able to dismiss this error? Otherwise it just hangs out until the plugin is updated to support my WooCommerce version (good thing always on top of compatibility releases!!)

packages/php/compat-checker/src/Checks/WCCompatibility.php
```diff
@@ -0,0 +1,389 @@
...
+			/* translators: Placeholders: %1$s - plugin name, %2$s - WooCommerce version number, %3$s - opening <a> HTML link tag, %4$s - closing </a> HTML link tag */
+			esc_html__( 'Heads up! %1$s will soon discontinue support for WooCommerce %2$s. Please %3$supdate WooCommerce%4$s to take advantage of the latest updates and features.', 'woogrow-compat-checker' ),
+			'<strong>' . $plugin_name . '</strong>',
+			$current_wc_version,
+			'<a href="' . esc_url( admin_url( 'update-core.php' ) ) . '">',
+			'</a>'
+		);
+
+		$this->add_admin_notice(
+			'woocommerce-upgrade-recommendation',
```
layoutd, line 369: This seems like it should be dismissible or at least snoozeable - if the merchant can't update WooCommerce right now for whatever reason, it's possible that all of a sudden they'll see multiple nag warning notices.

packages/php/compat-checker/src/Checks/WPCompatibility.php
```diff
@@ -0,0 +1,59 @@
...
+ */
+
+namespace Automattic\WooCommerce\Grow\Tools\CompatChecker\v0_0_1\Checks;
+
+defined( 'ABSPATH' ) || exit;
+
+/**
+ * The WooCommerce Compatibility Check class.
+ */
+class WPCompatibility extends CompatCheck {
```
layoutd, line 15: ❓ I can imagine it would almost never be a problem IRL, but should there be a check for the WP requires at least header? Or do we let the `WC requires at least` set the WP version requirement too?

packages/php/compat-checker/src/Checks/WCCompatibility.php
```diff
@@ -0,0 +1,389 @@
...
+
+		$plugin_name        = $this->plugin_data['Name'];
+		$current_wc_version = $this->get_wc_version();
+
+		$message = sprintf(
+			/* translators: Placeholders: %1$s - plugin name, %2$s - WooCommerce version number, %3$s - opening <a> HTML link tag, %4$s - closing </a> HTML link tag */
+			esc_html__( 'Heads up! %1$s will soon discontinue support for WooCommerce %2$s. Please %3$supdate WooCommerce%4$s to take advantage of the latest updates and features.', 'woogrow-compat-checker' ),
+			'<strong>' . $plugin_name . '</strong>',
+			$current_wc_version,
+			'<a href="' . esc_url( admin_url( 'update-core.php' ) ) . '">',
```
layoutd: This URL is better, it just takes the merchant to the screen where they can decide whether to update WooCommerce or not. 
💅🏻  Maybe add `#update-plugins-table` to the URL to jump straight to upgradeable Plugins?

layoutd (changes requested):
😮  This is very cool! <u>Thank you</u> for the extremely detailed testing instructions. The tests worked well with Brands and Product Filters, although I did discover that I had to delete the transients repeatedly once I started changing the versions in the header.

I added a handful of comments, mostly about links and notice settings.

ibndawood:
Thank you, @layoutd, for taking the time to review this PR. I'll go through your feedback and shall add my comments/fixes.

packages/php/compat-checker/src/Checker.php
ibndawood: The plugin file version is part of the `$transient_key`. The updated plugin will have a new `$transient_key` because of the version change. (props to @mikkamp for the idea). Discussion about the transient expiration time here: https://github.com/woocommerce/grow/pull/79#issuecomment-1695410469

packages/php/compat-checker/src/Checker.php
ibndawood: Agree with your observation. I've implemented the suggestion here: https://github.com/woocommerce/grow/pull/79/commits/f9937523235e2319f5e5800bc512f0f6e3d0a23a

packages/php/compat-checker/src/Checks/WPCompatibility.php
ibndawood: I had added this check earlier but while testing I noticed that WordPress core has the `WP requires at least` check internally and deactivates the incompatible plugins. So I've removed our redundant implementation. Thank you for checking this.

packages/php/compat-checker/src/Checker.php
layoutd: Ah, right! I read through the conversation thread before the review, somehow that part didn't stick  in my brain.

packages/php/compat-checker/src/Checks/WPCompatibility.php
layoutd: 👍🏻  That makes sense. It's a pretty basic check, so I suspected there must be some reason to leave it out.

packages/php/compat-checker/src/Checks/WCCompatibility.php
ibndawood: Thank you for this suggestion. I've added the anchor jump to the URL fragment in this commit: https://github.com/woocommerce/grow/pull/79/commits/394a108d78dcbb8f3c77b6861d0b6a2828ffd320

packages/php/compat-checker/src/Checks/WCCompatibility.php
ibndawood: This is an excellent suggestion. I've replaced the update URL with the link to the update core URL: https://github.com/woocommerce/grow/pull/79/commits/efbbd1954ce28794fb73b27bfff081ead450e723

packages/php/compat-checker/src/Checks/WCCompatibility.php
ibndawood: This is a good catch. Thank you. I've added padding in this commit: https://github.com/woocommerce/grow/pull/79/commits/115cbc9497f85248c0c07d9cf6d88fda0fba0902

packages/php/compat-checker/src/Checks/WCCompatibility.php
ibndawood: I've made the warning notice dismissible in this commit: https://github.com/woocommerce/grow/pull/79/commits/c8c50e043763f275f1e5901ba2975b0aab2416f1 

The `is-dismissible` applies only on the current screen. To maintain the notice state, we'll have to store the notice state in the user meta data. Do you want me to implement that?

packages/php/compat-checker/src/Checks/WCCompatibility.php
ibndawood: Same as the comment here: https://github.com/woocommerce/grow/pull/79#discussion_r1333000743

I've added the `is-dismissible` class to the warnings but that applies only to the current screen.

ibndawood:
Hello again, @layoutd. Thank you for taking the time to review this PR. I've made the changes you suggested. This is ready for another round of review. I've also added a question related to the dismiss notice. Please let me know what you think.

packages/php/compat-checker/src/Checks/WCCompatibility.php
layoutd: Hmm ... I guess for WooCommerce it's not particularly critical because our extensions _should_ always be up to date with WooCommerce. We could either make it dismissible until the next WooCommerce update, or leave as is and then change it if we see that our extensions are falling behind on compatibility (+1 technical debt)

layoutd (approved):
LGTM - the changes work well and I think this is OK as is. I still have a few more comments though (😳)

- There's a bit of a hiccup in the "Install WooCommerce" link process, where the prompt to install the plugin appears on the page where I just installed the plugin:

I wonder if that check could be skipped on that page in particular? But it's not a big deal since it ends up resolving itself, just could be a bit confusing.
- I added some notes to the testing instructions as a reminder that that the compatibility  transient needs to be erased for version testing.
- I wonder if this text should just be `WooCommerce Brands version 1.6.58 …` or `WooCommerce Brands - version 1.6.58 … `.
image

ibndawood:
> I wonder if that check could be skipped on that page in particular? But it's not a big deal since it ends up resolving itself, just could be a bit confusing.

I agree. I'll fix this.

> I added some notes to the testing instructions as a reminder that that the compatibility transient needs to be erased for version testing.

Thank you. 

> I wonder if this text should just be `WooCommerce Brands version 1.6.58` … or `WooCommerce Brands - version 1.6.58` … 

I'd go with `WooCommerce Brands - version 1.6.58`.

ibndawood:
Hey @layoutd, I've addressed the below comment in https://github.com/woocommerce/grow/pull/79/commits/a8579b7dfb444d3a1fee0650797d0e5645936825:

> I wonder if that check could be skipped on that page in particular? But it's not a big deal since it ends up resolving itself, just could be a bit confusing.

In https://github.com/woocommerce/grow/pull/79/commits/e7bf974d85ab5e2fbf11bb4065f381ee3223d0a1, I've set the text in the notice like this: `WooCommerce Brands - version 1.6.58`.

I'd like you to take a look at these two changes and let me know your feedback.

layoutd (approved):
Nice! Both "issues" look like they've been solved, thanks for humoring me @ibndawood!

# :shipit: 

image
image

ibndawood:
Thank you, @layoutd, for your time to review the final changes. I'll merge this PR to the `trunk`.

PR #81, opened 2023-09-27
Add actions and workflows for QIT tests

tomalec:

This PR adds workflows and actions to be used here and in extensions' repos.

- `run-qit-annotate` action is a basic building block. It takes test `type`, `extension` name, and optionally `extension-file`, to run a `qit` test, wait for results, and propagate them to annotations and output. It also takes a few optional params for customization: 
   - `options` - so a user can provide additional options to `qit` command, like `--optional_features=hpos`, or `--woocommerce_version` (currently broken https://github.com/woocommerce/qit-cli/issues/81)
   - `wait` to wait for the results, otherwise it just fires the test run, and returns its id for further checks
   - `ignore-fail` a flag to consider a failed test "ok" thing, for example, to proceed with other test types
- `run-qit-extension` is another action to run all test types for a given extension. It takes `qit-partner-*` secrets, set of `run-qit-annotate` inputs (`extension`, `wait`, `ignore-fail`, `options`) and `test-*` flag for each type to run. It also takes `version` input:
   - `latest` will test the latest publicly released version,
   - `dev` will make the action look up for an `{extension}.zip` file in the root of the repository under the `gha-dev-build` tag,
   - `local` will look up for an `{extension}.zip` artifact in the current workflow.
- `.github/workflows/run-qit-all.yml` local manual workflow that uses the above to be run on demand to test all the extensions at once. Currently, it's configured to test Automata extensions.

To make those workflows run, three secrets are required to be set up:
- `BOT_GH_TOKEN`
- `QIT_PARTNER_USER`
- `QIT_PARTNER_SECRET`

### Screenshots:
image

image

image

#### Manual workflow

1. Fork the repo
2. Setup secrets. Links to secrets are at Pe98dy-Ir-p2 
3. Merge this branch to the main one /`trunk`
4. Run the manual workflow at  https://github.com/woocommerce/grow/actions/workflows/run-qit-all.yml

#### Reusable action
See https://github.com/woocommerce/google-listings-and-ads/pull/2114

1. Use a repo that builds a QIT-compatible extension
2. make sure you have another workflow that builds it and publishes the artifact if you want to test `local` version. You can test `latest` and `dev` without this step.
3. Use this workflow
```yml      
jobs:
  build:
    name: Build extension
    uses: ./.github/workflows/build.yml
    secrets: inherit
  qit-tests:
    name: Run QIT Tests
    runs-on: ubuntu-20.04
    needs: build
    steps:
      - name: Download artifact
        uses: actions/download-artifact@v3
        with:
          name: automatewoo.zip
      - name: Delegate QIT Tests
        uses: woocommerce/grow/run-qit-extension@add/qit-workflows-test-build
        with:
          qit-partner-user: ${{ secrets.QIT_PARTNER_USER }}
          qit-partner-secret: ${{ secrets.QIT_PARTNER_SECRET }}
          version: local
          wait: true
          extension: 'automatewoo'
          test-activation: true
          test-security: true
          ignore-fail: true
          options: '--optional_features=hpos'
```

### Additional details:

1. Manual workflow can be tested only once merged to the main branch. So you can try it on your local fork.
5. ~Due to the above, for show-case purposes, I run the workflow on the versions matrix for each push to this branch. https://github.com/woocommerce/grow/actions/runs/6331342166~
6. ~We publish here a reusable workflow. I was thinking of pushing it to `github-actions` folder. So, we will have it under common release flow and location, with more convenient version tags. But I'm not sure, as this is "workflow" not "action". Maybe we can store it in a location like `.github/workflows/reusable` or `packages/workflows` and preserve that folder in the github-actions release?~ (edit: as per [the discussion below](https://github.com/woocommerce/grow/pull/81#discussion_r1342523227), we decided to go with a another composite action)
4. :warning: before merging, we need to update references from `woocommerce/grow/run-qit-*@add/qit-workflows-test-build` to `woocommerce/grow/run-qit-annotate@actions-v1`

### Changelog entry

> Add QIT workflows and actions.


.github/workflows/run-qit-extension.yml
```diff
@@ -0,0 +1,154 @@
...
+                repository: "woocommerce/${{ inputs.extension }}"
+                tag: 'gha-dev-build'
+                fileName: ${{ env.dev_build }}
+                token: ${{ secrets.BOT_GH_TOKEN }}
+
+            - name: Download artifact
+              if: ${{ inputs.version == 'local' }}
+              uses: actions/download-artifact@v3
+              with:
+                name: ${{ env.dev_build }}
```
eason9487: If I guess it right, this step is meant to be used in a way like [this line](https://github.com/woocommerce/google-listings-and-ads/pull/2114/files#diff-0550f940c28a3fbd24937bad01914ede672b538aff10de709e999759ccf4f9dfR54)?

A concern is the workflow files in `.github/workflows` directory of this repo won't be included in the `actions-v1` releases. As a result, it would become difficult to ensure version compatibility between them.

To run QIT tests with a local build, I would suggest leaving the processing of preparing a local extension zip file on the action user side. For example, GLA could either build a zip file within its workflow or run the `actions/download-artifact` step on its own, and this action only needs the local file path as input.

packages/github-actions/actions/run-qit-annotate/README.md
```diff
@@ -0,0 +1,56 @@
...
+        timeout-minutes: 5
+        with:
+          type: security
+          extension: automatewoo
+          options: '--optional_features=hpos'
+
+      - name: Echo status
+        shell: bash
+        run: echo ${{ jobs.security-test.outputs.status }}
+
```
eason9487: Mixed indentation could be removed.

```suggestion
```

eason9487 (commented):
Thanks for the work.

The current way of transferring a local extension zip might not be easy to handle in the future in terms of version compatibility. Wonder if it would be better to leave the processing to the action user side.

Another concern is about the security reports. Since this repo is public, the security test reports for the private repos such as AW will also reveal the report URLs after running `.github/workflows/run-qit-all.yml`.

For example, it reveals the "****" POST parameter is not sanitized. (I replaced the actual info with `****` via DevTool)

image

I'd like to clarify whether this is appropriate?

.github/workflows/run-qit-extension.yml
tomalec: Yes, this is the way to use it.


> A concern is the workflow files in `.github/workflows` directory of this repo won't be included in the actions-v1 releases. As a result, it would become difficult to ensure version compatibility between them.

I completely agree with that concern. That's why in the OP I mentioned:
> I was thinking of pushing it to `github-actions` folder. So, we will have it under common release flow and location, with more convenient version tags. But I'm not sure, as this is "workflow" not "action". Maybe we can store it in a location like `.github/workflows/reusable` or `packages/workflows` and preserve that folder in the github-actions release?

I would appreciate your opinion here. I'd say having separate release flows for workflows and separate for actions is too much. I would choose one of those approaches:

- Consider our _[github] <code><strong>actions</strong></code>_ tag actually mean _"github **actions & workflows**"_, then
   1. consider workflows and atomic actions completely equal, and make consumers distinguish what is what:
   ```yml
   jobs:
     build:
       runs-on: ubuntu-latest
       steps:
        - name: Reuse action
          uses: woocommerce/grow/prepare-php@actions-v1 # Action - should be used in `jobs.*.steps.uses`
     runQIT:
       name: Reuse workflow
       uses: woocommerce/grow/run-qit-extension.yml@actions-v1 # Workflow - should be used in `jobs.*.uses`
       needs: build
   ```
   2. put workflows and actions into distinct folders (so maintainers and consumers can easily distinguish what is what)
   ```yml
   jobs:
     build:
       runs-on: ubuntu-latest
       steps:
        - name: Reuse action
          uses: woocommerce/grow/action/prepare-php@actions-v1
     runQIT:
       name: Reuse workflow
       uses: woocommerce/grow/workflow/run-qit-extension.yml@actions-v1
       needs: build
   ```

> To run QIT tests with a local build, I would suggest leaving the processing of preparing a local extension zip file on the action user side. For example, GLA could either build a zip file within its workflow or run the actions/download-artifact step on its own, and this action only needs the local file path as input.

The problem here is that. This is a workflow (`jobs.*.uses`), not an atomic action (`jobs.*.steps.uses`). This workflow runs as a separate job, from any other workflow or action. So the only way to share a file (I'm aware of) between this workflow and another (GLA's, or other extension's) workflow is to use artifacts. Then this workflow needs to download them itself. Otherwise, how would GLA's job push a file to this job?

tomalec:
Thank you for the deep review :) 

> The current way of transferring a local extension zip might not be easy to handle in the future in terms of version compatibility. Wonder if it would be better to leave the processing to the action user side.

Could you elaborate on that? What challenges do you see with that approach in regard to version compatibility? BTW, by "version compatibility" do you mean "testing extension's compatibility with WC/WP version", or something else?
How would you make it so that the user would push the extension zip to another workflow?

tomalec:
> Another concern is about the security reports. Since this repo is public, the security test reports for the private repos such as AW will also reveal the report URLs after running `.github/workflows/run-qit-all.yml`.
> ...
> I'd like to clarify whether this is appropriate?

Thanks for catching that! I confirmed with Masamune Team, that test results should be considered private to the direction of the extension vendor. So for 3PD QIT access, a vendor can see theirs and only their extension results. And it's up to the vendor what they do with the results.

I posted a comment on the project thread so we could discuss or approach and policy peeuvX-QJ-p2#comment-1215

I'll look for ways to hide those results

.github/workflows/run-qit-extension.yml
eason9487: The version compatibility concern in https://github.com/woocommerce/grow/pull/81#issuecomment-1743623849 is the same as this thread.

> > The current way of transferring a local extension zip might not be easy to handle in the future in terms of version compatibility. Wonder if it would be better to leave the processing to the action user side.
> 
> Could you elaborate on that?

Sorry, I omitted a large portion of explanation. The idea is to transform the `run-qit-extension.yml` workflow into a composite action and it involves a degree of change that is not quite easily stated in words. I created a demo in the repo and one in GLA to illustrate how it works.

### [Changes in this repo](https://github.com/woocommerce/grow/commit/d7e552ea0e420901c3b5baff6629a5af04be16f9)

- Based on the `run-qit-extension` workflow, create a composite `run-qit-extension` action.
- Remove the "Download artifact" step.
- Add `qit-partner-user` and `qit-partner-secret` inputs to forward the original secret variables.
- Remove all `timeout-minutes` as they are not supported in the composite action. Probably it would be better to let the action user specify the timeout.
- I didn't transform the `outputs.statuses` as it would need more time or the composite action might not support the same syntax.
   - https://github.com/woocommerce/grow/blob/a8f23bae23cbe0e896189ec561099897622e4ac1/.github/workflows/run-qit-extension.yml#L52-L55
   - https://github.com/woocommerce/grow/blob/a8f23bae23cbe0e896189ec561099897622e4ac1/.github/workflows/run-qit-extension.yml#L65-L66

### [Changes in GLA](https://github.com/woocommerce/google-listings-and-ads/commit/b0ed7ea612eddc38e8f627b0af3d67f12b1dc045)

- Add the "Download artifact" step.
- Use the `run-qit-extension` action instead.

---

The test run: https://github.com/woocommerce/google-listings-and-ads/actions/runs/6391261008

I didn't continue to make further changes. If adopt this method, the `.github/workflows/run-qit-all.yml` workflow should be able to use `run-qit-extension` action as well, and the `.github/workflows/run-qit-extension.yml` workflow would be no longer in needed.

In this way, the action user sides won't need to use this repo's workflow directly, and this repo doesn't need to add them to any release and versioning flow.

Another advantage is that the action user side wouldn't have to upload artifacts if their action builds a local zip file and uses this action in the same job. For example:

```yaml
# ... omitted

jobs:
  qit-tests:
    name: Run QIT Tests
    runs-on: ubuntu-20.04
    steps:
      - name: Checkout repository
        uses: actions/checkout@v3

      # ... omitted

      - name: Build production bundle
        run: |
          npm run build

      - name: Delegate QIT Tests
        uses: woocommerce/grow/run-qit-extension@actions-v1
        with:
          qit-partner-user: ${{ secrets.QIT_PARTNER_USER }}
          qit-partner-secret: ${{ secrets.QIT_PARTNER_SECRET }}
          version: 'local'
          extension: 'google-listings-and-ads'
          test-security: 'true'
```

eason9487:
Just noticed the indentation of some .yml files is mixed with 2-space and 4-space units. It would be better to make them consistent.

packages/github-actions/actions/run-qit-annotate/action.yml
```diff
@@ -0,0 +1,94 @@
...
+      env:
+        # If the custom build is provided, pass it to the `zip` param.
+        zip: ${{ inputs.extension-file &&  format('--zip={0}', inputs.extension-file) || '' }}
+        report_file: ${{ inputs.results-filename || format( '{0}/{1}/{2}/{3}.txt', github.workspace, inputs.results-folder, inputs.extension, inputs.type ) }}
+        status: 0
+      run: |
+        ./vendor/bin/qit run:${{ inputs.type }} ${{ inputs.extension }} \
+          $zip \
+          ${{ inputs.options }} \
+          --wait \
```
eason9487: > [...], wait for results, and propagate them to annotations and output.

Maybe removing the `--wait` parameter and leaving the use of this option up to users would be better.

For example, if a user only wants to delegate QIT tests through this action and view report pages later, they probably don't expect to wait for QIT tests to finish running, it's also beneficial to save billing time for private repos.

.github/workflows/run-qit-extension.yml
tomalec: Thank you for the detailed explanations and the new branch! 
I can see it works, and I appreciate how it addresses the following:

> this repo doesn't need to add them [workflows] to any release and versioning flow.

However, I'm not sure if I see all the benefits of 
>  the action user sides won't need to use this repo's workflow directly

- I get that accessing the workflow directly is not great, and I'm happy to improve it. But I think we can to it also otherwise
- Using action or workflow from code perspective is not much longer


The biggest difference I see is granularity. Having `qit` as an atomic action, users could compose it within their workflows.
But I think it also has the downsides:
- the log is less collapsible and divided.:
as a workflow: 
image
as an action
image

- I was actually thinking of unifying the workflow across our repos so we will not have to implement a workflow and compose actions in every repo but just use the shared entire workflow.


Especially if it comes to the build, I think it is even desirable to have it as a separate workflow that uploads an artifact. So, other workflows and actions running for the same commit could reuse that work, reducing the overall cost and time.
See https://github.com/woocommerce/google-listings-and-ads/pull/2114#discussion_r1341025517

tomalec:
Speaking of reusable entire workflows, I think we could benefit from having them for other cases like  [.github/workflows/js-css-linting.yml](https://github.com/woocommerce/automatewoo-birthdays/pull/91/files#diff-461537f663f8b1c859af34a40d21118123ea130269a3743becbd7110d7866b94)

.github/workflows/run-qit-extension.yml
eason9487: > * I was actually thinking of unifying the workflow across our repos so we will not have to implement a workflow and compose actions in every repo but just use the shared entire workflow.

From the perspective of this repo, I would suppose its granularity prefers a [custom action](https://docs.github.com/en/actions/creating-actions/about-custom-actions) over a [reusable workflow](https://docs.github.com/en/actions/using-workflows/reusing-workflows#creating-a-reusable-workflow) for flexibility in composition and configuration and fewer limitations.

If an automation requirement is highly repetitive or identical across several repos, then a reusable workflow, which might still composite some custom actions within this repo, would be appropriate.

> Especially if it comes to the build, I think it is even desirable to have it as a separate workflow that uploads an artifact. So, other workflows and actions running for the same commit could reuse that work, reducing the overall cost and time. [...]

From a practicable point of view, I find it rather difficult to imagine whether it would be easier to do so.

Take GLA as an example, given it has a reusable workflow for building a zip bundle, and each automation of "Bundle watch", "Publish dev build", "E2E test", and "QIT" need a build file as input.

Unless all these automations are in the same workflow, each automation will be grouped into standalone workflows with the same execution conditions, and then build a zip on its own. Because artifact access through `actions/upload-artifact` and `actions/download-artifact` is limited within the same workflow run. In this case, the same commit doesn't use the "same" work result (zip bundle), only the extracted build workflow is reused.

However, having all of them in the same workflow run seems to make that workflow configuration very complex.

eason9487: My main concern is still how the release versions of these reusable workflows in the `.github/workflows` directory will proceed if go with the current solution.

The unintuitive understanding of the artifact upload/download split between the use-end repo and this repo is also a concern, but if this complexity trade-off is necessary, then it won't be a blocker.

> I'd say having separate release flows for workflows and separate for actions is too much.

Agree with this.

> I would choose one of those approaches: [...]

GitHub docs say the syntax of [`jobs.<job_id>.uses`](https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#jobsjob_iduses) is `{owner}/{repo}/.github/workflows/{filename}@{ref}` and doesn't clarify if the `.github/workflows/` part can be changed.

Either yes or no, I lean toward leaving reusable workflows under the original directory and keeping custom actions the same as it's:

```yml
jobs:
  build:
    runs-on: ubuntu-latest
    steps:
     - name: Reuse action
       uses: woocommerce/grow/prepare-php@actions-v1

  runQIT:
    name: Reuse workflow
    uses: woocommerce/grow/.github/workflows/run-qit-extension.yml@actions-v1
    needs: build
```

The main reason is that each of them conforms to the syntax described in GitHub docs respectively.
- [`jobs.<job_id>.steps[*].uses`](https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#jobsjob_idstepsuses)
- [`jobs.<job_id>.uses`](https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#jobsjob_iduses)

tomalec: I merged your proposed changes and used action as you suggested.
We can get back to shared workflows if needed in some following PRs, exploring possibilities and flow optimizations there.

tomalec:
Speaking of hiding fragile data. Now, we echo only parsed json output, that contains `{test_run_id, status, test_summary}`, like 
```js
{
  "test_run_id": 4554116,
  "status": "failed",
  "test_summary": "Errors: 0, File Errors: 86"
}
```
I believe it's the right balance between the developer's convenience and exposing vulnerabilities.. The annotation also contains a ready CLI command to check for more, like:
image


So, we no longer expose the results URL with a private token, but only test_run_id. IMHO, it should not be considered private, as it's just a consecutive number, so anybody could guess that. Then, it's up to QIT CLI to guard against unauthorized `qit get ...`

tomalec:
@eason9487 Thank you once again for the detailed review and great insights. Also, thank you for the "GitHub Actions - Create Test Build" - it made development and testing much easier.

I addressed all your comments, and the PR is ready for another round of review.

packages/github-actions/README.md
```diff
@@ -20,6 +20,7 @@ Custom GitHub actions that help to composite GitHub workflows across the repos m
 - [`prepare-node`](actions/prepare-node) - Set up Node.js with a specific version, load npm cache, install Node dependencies
 - [`prepare-php`](actions/prepare-php) - Set up PHP with a specific version and tools, load Composer cache, install Composer dependencies
 - [`publish-extension-dev-build`](actions/publish-extension-dev-build) - Publish extension development build
+- [`run-qit-annotate`](actions/run-qit-annotate) - Runs QIT test and annotates the results
```
eason9487, line 23: The `run-qit-extension` action could be listed here as well.

packages/github-actions/actions/run-qit-annotate/README.md
```diff
@@ -0,0 +1,50 @@
...
+      QIT_DISABLE_ONBOARDING: yes
+    steps:
+      - name: Install QIT via composer
+        run: composer require woocommerce/qit-cli
+
+      - name: Add Partner
+        run: |
+          ./vendor/bin/qit partner:add \
+            --user='${{ secrets.QIT_PARTNER_USER }}' \
+            --application_password='${{ secrets.QIT_PARTNER_SECRET }}'
```
eason9487: image

Noticed this deprecation message on the GL&A side. Looks like this could be changed to use [`qit_token`](https://github.com/woocommerce/qit-cli/blob/0.4.0/src/src/Commands/Partner/AddPartner.php#L42-L43).

packages/github-actions/actions/run-qit-extension/action.yml
```diff
@@ -0,0 +1,151 @@
...
+    - name: Install QIT via composer
+      shell: bash
+      run: composer require woocommerce/qit-cli
+
+    - name: Add Partner
+      shell: bash
+      run: |
+        ./vendor/bin/qit partner:add \
+          --user='${{ inputs.qit-partner-user }}' \
+          --application_password='${{ inputs.qit-partner-secret }}'
```
eason9487: Same as the above comment.

eason9487 (approved):
Thanks for the great work! LGTM.

Left two minor comments and they are not blockers.

PR #86, opened 2023-10-22
Add `nvm use` and auto-branching

message-dimke:

This PR extends `release` command with additional parameters
```
--nvm_use        - used to tell `release` to use `nvm use` per project basis.
--default_branch - used to pass project's default repo branch name into `release` command.
```
PR also adds few commands to be using as complementary commands to the overwritten `release` command:
```
simulate - used to run simulation mode on the new release command.
branch   - used to add a dialog for release branch creation.
```
Added set of tools for `Git`, `Nvm` and `Product`.

It is now possible to run the `release` and `simulate` commands using new parameters and see some changes in the command workflows.
```bash
./bin/wr simulate --nvm_use --default_branch develop --generate_changelog --wc_tested 8.3 --wp_tested 6.4 https://github.com/woocommerce/pinterest-for-woocommerce/tree/testrelease/1.3.13
```
For example if you provide a GitHub url using a branch name which does not exist on the repository, you will see a prompt to create it and release from there.

Terminal

It is also possible to skip release branch creation and release from the default which is set to `develop` by default. All you need is to refuse release branch creation.

Terminal

The invisible part of changes introduced with this PR is `nvm use` command to run for every releasing product being inside the corresponding product directories to switch the release process to a proper `node` and `npm` versions. If `--nvm_use` flag is provided into `release` command, the additional logic will get involved into `build` process, if not - then the `release` will flow as usual.

1. Checkout `add-nvm-and-autobranching` branch.
2. Open `release.txt` in the root and add some lines to it.
```txt
pinterest-for-woocommerce	1.3.13	testrelease/1.3.13
woocommerce-brands	1.6.62	testrelease/1.6.62
```
**Note:** I am using Pinterest and Brands here because both projects have to use their own `node` version and w/o `nvm use` the build will fail. Pinterest can build with `node 14` when Brands require `node 16` for release.

You can provide any nonsense instead of a release branch names to check how branch prompt is working.
3. Run bulk command
```bash
./bin/wr bulk simulate --generate_changelog --wc_tested 8.3 --wp_tested 6.4
```
**Note:** I am using `bulk` command here as a single command to work with `release.txt`.  Also you may notice `--wc_tested` and `--wp_tested` flags, those are also here to check version bump still functional.
4. While releasing wait for a branch creation prompt to appear.
5. Choosing `y` will start a branch creation process. Branch will be created from a default one and pushed to remote. The push is requires for a `changelog` command to generate its changelog part.
6. Choosing `N` will drop release branch creation process and prompt to release from a default branch which is, if rejected, will terminate the release process.
7. When running the simulation every release branch, except the default, must be cleaned up automatically at the end of the process.
8. All the steps are followed with the corresponding console messages/notifications/errors.
### Changelog entry

> Dev - Adding auto-release branching.
> Dev - Adding `nvm use` on product basis.


extensions.json
```diff
@@ -14,19 +14,22 @@
 		},
 		{
 			"name": "Points & Rewards",
-			"repoSlug": "woocommerce-points-and-rewards"
+			"repoSlug": "woocommerce-points-and-rewards",
+			"defaultBranch": "develop"
```
JPry: Since `develop` is now the default branch for our repos, we should instead change the `defaultBranch` property at the top of the file, and then we can skip all of these definitions.

packages/php/woorelease-extension/src/Utils/Nvm.php
```diff
@@ -0,0 +1,47 @@
...
+class Nvm {
+
+	/**
+	 * Checks if `nvm` is installed.
+	 *
+	 * @return bool True if `nvm` is installed, false otherwise.
+	 */
+	public static function is_nvm_exists() {
+		try {
+			Utils::exec_sprintf( 'nvm -v' );
```
JPry:
```suggestion
			Utils::exec_sprintf( 'which -s nvm' );
```

packages/php/woorelease-extension/src/Commands/Release.php
```diff
@@ -0,0 +1,277 @@
...
+ * Class for implementing the release command.
+ */
+class Release extends WooReleaseRelease {
+
+	protected static $defaultName = 'release';
+
+	public function __construct( string $name = null ) {
+		parent::__construct( $name );
+	}
+
```
JPry: We can omit all of this since we're not actually making changes.

```suggestion
```

.gitignore
```diff
@@ -8,4 +8,5 @@ vendor/
 *.zip
 
 # IDE files
-.vscode/*
\ No newline at end of file
+.vscode/*
+.idea
```
JPry: This should be part of your global `.gitignore`. 

```suggestion
```

bin/wr
```diff
@@ -3,6 +3,7 @@
 
 use Automattic\WooCommerce\Grow\WR\Commands\Bulk;
 use Automattic\WooCommerce\Grow\WR\Commands\VersionReplace;
+use Automattic\WooCommerce\Grow\WR\Commands\Release;
```
JPry: Minor: it's nice to keep them in alphabetical order 🙂 

```suggestion
use Automattic\WooCommerce\Grow\WR\Commands\Release;
use Automattic\WooCommerce\Grow\WR\Commands\VersionReplace;
```

packages/php/woorelease-extension/src/Utils/Git.php
```diff
@@ -0,0 +1,46 @@
...
+
+	/**
+	 * Checks if a branch exists in a remote repository.
+	 *
+	 * @param string $repository_url The remote repository URL.
+	 * @param string $branch         The branch name.
+	 *
+	 * @return bool True if the branch exists, false otherwise.
+	 */
+	public static function is_branch_exists( $repository_url, $branch ) {
```
JPry: Changing the name helps this to read better.

```suggestion
	public static function does_branch_exist( $repository_url, $branch ) {
```

packages/php/woorelease-extension/src/Utils/Git.php
```diff
@@ -0,0 +1,46 @@
...
+		// --exit-code returns 0 if the branch exists, 2 otherwise.
+		try {
+			Utils::exec_sprintf( 'git ls-remote --exit-code --heads %s %s', $repository_url, $branch );
+			return true;
+		} catch ( \Exception $e ) {
+			return false;
+		}
+	}
+
+	public static function create_branch( $repository_url, $branch ) {
```
JPry: The `$repository_url` parameter isn't used, and we're missing a docblock.

packages/php/woorelease-extension/src/Utils/Nvm.php
```diff
@@ -0,0 +1,47 @@
...
+ * Check if `nvm` CLI tool is installed.
+ */
+class Nvm {
+
+	/**
+	 * Checks if `nvm` is installed.
+	 *
+	 * @return bool True if `nvm` is installed, false otherwise.
+	 */
+	public static function is_nvm_exists() {
```
JPry: Changing the name helps this to read better.

```suggestion
	public static function does_nvm_exist() {
```

JPry (commented):
I know this is progress, and I'm not able to dedicate time to a full review right now, but I just wanted to leave a few comments.

packages/php/woorelease-extension/src/Utils/Nvm.php
message-dimke: Hey, @JPry ! I found `-s` is not an option on ubuntu for example. There is some inconsistency between operational systems. Do we still want it?
https://linux.die.net/man/1/which
https://ss64.com/osx/which.html

message-dimke: Also, it is interesting but `which -s nvm` was working for me but then stopped. Running the command now gives me a function nvm() {} output into my console. I googled this:
```
The nvm command is not a file but rather a shell function.
The source ~/.nvm/nvm.sh adds these functions to your current shell. And because these commands are not files, they don't show up when you which nvm.
Looking in the .nvm/nvm.sh file, you can see a nvm() {...} function defined that provides that functionality.
```

message-dimke: Also in Terminal I have zsh and exec uses sh. That is the reason I see some inconsistencies in commands when running from both places.

message-dimke: Closing the conversation because of another way of solving the issue introduced with `/bin/build` and `/bin/nvm` shell scripts.

rawdreeg (approved):
Sorry for the delay. This has been in my to-do list for a while lol.

Looks good, all the previous feedback and nits are now fixed, and I couldn't find any other issue.

While testing this I encountered #99 and I submitted a PR fixing it (#100). So I'll approve this!

PR #87, opened 2023-10-24
Use different objects of the CompatCheck class based on plugin basename

ibndawood:

The `CompatCheck` class creates instances based on the object. This PR uses the `plugin_basename` to create more unique objects. Adding this allows the compat checker to be used with multiple plugins.

1. Setup Compat Checker in two WooCommerce extension by following the [Getting Started](https://github.com/woocommerce/grow/tree/trunk/packages/php/compat-checker#getting-started) section. Let the two extensions be WooCommerce Points and Rewards and WooCommerce Coupon Campaigns.
2. Run all the tests in the Compat Checker PR from here: https://github.com/woocommerce/grow/pull/79 with both the extensions activated.
3. Before PR: Only admin notice from one plugin is displayed because the `CompatCheck` class generates only one object.
4. After PR: Notices from both the plugins are displayed.

### Screenshots

image

### Changelog entry

> Add - Different instances of CompatCheck based on plugin basename.

### DIFF
--- a/packages/php/compat-checker/src/Checker.php
+++ b/packages/php/compat-checker/src/Checker.php
@@ -71,14 +71,15 @@ class Checker {
 	 * @return bool
 	 */
 	public function is_compatible( $plugin_file_path, $file_version ) {
-		$checks      = array(
+		$checks          = array(
 			WPCompatibility::class,
 			WCCompatibility::class,
 		);
-		$plugin_data = $this->get_plugin_data( $plugin_file_path, $file_version );
+		$plugin_data     = $this->get_plugin_data( $plugin_file_path, $file_version );
+		$plugin_basename = plugin_basename( $plugin_file_path );
 
 		foreach ( $checks as $compatibility ) {
-			if ( ! $compatibility::instance()->is_compatible( $plugin_data ) ) {
+			if ( ! $compatibility::instance( $plugin_basename )->is_compatible( $plugin_data ) ) {
 				return false;
 			}
 		}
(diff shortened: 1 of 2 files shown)


layoutd (approved):
Nice! This works well and successfully separates the error messages for different extensions. 👏🏻 

One note is that in Points & Rewards, finding exactly where to run the checks was a bit tricky:
- `plugins_loaded` action is hooked to `woocommerce_points_and_rewards_init` on [L799](https://github.com/woocommerce/woocommerce-points-and-rewards/blob/0e8c77e65cda7cf3ea86b2a9c63338da1a81fb56/woocommerce-points-and-rewards.php#L799)
- However, the subsequent `woocommerce_blocks_loaded` hook is executed independently of `woocommerce_points_and_rewards_init`, and initializes the `WC_Points_Rewards_Integration` class
- This throws a class not found error because if the compat checker is placed at the start of `woocommerce_points_and_rewards_init`, the [`includes()` method is never invoked](https://github.com/woocommerce/woocommerce-points-and-rewards/blob/0e8c77e65cda7cf3ea86b2a9c63338da1a81fb56/woocommerce-points-and-rewards.php#L235), so the associated files are never included. In the end, I added the `Checker` after initializing the class:
```php
$GLOBALS['wc_points_rewards'] = WC_Points_Rewards::instance();
if ( ! Checker::instance()->is_compatible( __FILE__, WC_POINTS_REWARDS_VERSION ) ) {
	return;
}

ibndawood:
Thank you, @layoutd, for taking the time to review this PR. Also, excellent observation about the integration in `WooCommerce Points and Rewards`. I'll compare these findings with the WooCommerce Points and Rewards checker implementation PR.

PR #105, opened 2024-04-17
Adjust the publishing workflow of `compat-checker` to run only when the changed files are related to it

eason9487:

Currently, [the Publish Compat Checker package workflow](https://github.com/woocommerce/grow/actions/workflows/github-action-publish-compat-checker.yml) also runs when unrelated changes are pushed to `trunk`.

This PR adjusts the workflow to run only for the related changes.
Docs reference: https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#onpushpull_requestpull_request_targetpathspaths-ignore

N/A as it can only be tested after merging and having subsequent changes onto `trunk`

### DIFF
--- a/.github/workflows/github-action-publish-compat-checker.yml
+++ b/.github/workflows/github-action-publish-compat-checker.yml
@@ -3,6 +3,11 @@ on:
   push:
     branches:
       - trunk
+    paths:
+      - "**.js"
+      - "**.scss"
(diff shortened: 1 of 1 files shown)


eason9487:
I will be skipping the code review process as it can be optional for a devs-facing only change.

Issue #108, opened 2024-04-18
Upgrade to Node.js v20


eason9487:
## Description

This repo currently uses Node.js v14, which has reached its end-of-life 2023-04-30.

#### JS package `generator-grow`

- [x] Upgrade for dependencies, generators, and tests ([#111](https://github.com/woocommerce/grow/pull/111))
- [x] Update the actions' version in the templates of JS package `generator-grow` ([#129](https://github.com/woocommerce/grow/pull/129))

#### JS package `jsdoc` and `tracking-jsdoc`

- [x] Upgrade for JS package `jsdoc` and `tracking-jsdoc` ([#110](https://github.com/woocommerce/grow/pull/110))

#### Package `github-actions`

- [x] Update docs ([#123](https://github.com/woocommerce/grow/pull/123))
- [x] Adjust the v1 to be deprecated ([#124](https://github.com/woocommerce/grow/pull/124))
- [x] The fundamental setup and related workflows ([#112](https://github.com/woocommerce/grow/pull/112))
- [x] `automerge-released-trunk` action ([#114](https://github.com/woocommerce/grow/pull/114))
- [x] `branch-label` action ([#122](https://github.com/woocommerce/grow/pull/122))
- [x] `coverage-report` action ([#114](https://github.com/woocommerce/grow/pull/114))
- [x] `eslint-annotation` action ([#117](https://github.com/woocommerce/grow/pull/117))
- [x] `get-plugin-releases` action ([#119](https://github.com/woocommerce/grow/pull/119))
- [x] `get-release-notes` action ([#112](https://github.com/woocommerce/grow/pull/112))
- [x] `hook-documentation` action ([#115](https://github.com/woocommerce/grow/pull/115))
- [x] `merge-trunk-develop-pr` action ([#121](https://github.com/woocommerce/grow/pull/121))
- [x] `phpcs-diff` action ([#116](https://github.com/woocommerce/grow/pull/116))
- [x] `prepare-extension-release` action ([#121](https://github.com/woocommerce/grow/pull/121))
- [x] `prepare-mysql` action ([#115](https://github.com/woocommerce/grow/pull/115))
- [x] `prepare-node` action ([#112](https://github.com/woocommerce/grow/pull/112))
- [x] `prepare-php` action ([#114](https://github.com/woocommerce/grow/pull/114))
- [x] `publish-extension-dev-build` action ([#121](https://github.com/woocommerce/grow/pull/121))
[...]

PR #110, opened 2024-04-19
Upgrade JS packages `jsdoc` and `tracking-jsdoc` to use Node.js v20

eason9487:

Part of #108

This PR:

- Use Node.js v20 to regenerate the package-lock.json of JS package `tracking-jsdoc`.
- Set minimum support for Node.js to v16.
- Add the 'node:' prefix to explicitly indicate it requires a Node.js core module.
- Avoid generating package-lock.json file after `tracking-jsdoc` installation when developing this repo.

💡 The `<repo_directory>/.nvmrc` will be updated in a separate PR.

1. Use Node.js v20
2. `cd packages/js/jsdoc`
3. `npm install`
4. `npm run dev:doc:tracking`
5. Check if the generated content in `<repo_directory>/samples/jsdoc/TRACKING.md` is the same
6. Repeat the test with Node.js v16

### DIFF
--- a/packages/js/jsdoc/bin/jsdoc.mjs
+++ b/packages/js/jsdoc/bin/jsdoc.mjs
@@ -3,7 +3,7 @@
  * External dependencies
  */
 import shell from 'shelljs';
-import path from 'path';
+import path from 'node:path';
 
 process.env.PATH +=
 	path.delimiter + path.join( process.cwd(), 'node_modules', '.bin' );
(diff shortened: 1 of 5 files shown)


martynmjones (approved):
Hey @eason9487, thanks for updating Node!

Tested locally with `20.5.0` / `16.20.2` and the generated content in TRACKING.md was identical so LGTM ✅

PR #115, opened 2024-04-24
Update the version in docs of the `hook-documentation`, `prepare-mysql`, and `run-qit-annotate` actions to v2

eason9487:

Part of #108

This PR updates docs for 3 actions that don't use Node.js or other actions:

- `hook-documentation`
- `prepare-mysql`
- `run-qit-annotate`

These actions won't have a difference with v1 but it would be nicer to update their major version to v2 together. 

Check if there are any relevant content is missing from the update.


eason9487:
I will be skipping the code review process as it can be optional for a devs-facing only change.

PR #119, opened 2024-04-25
Upgrade the `get-plugin-releases` action to use Node.js v20

eason9487:

Part of #108

This PR upgrades the `get-plugin-releases` action to use Node.js v20.
In addition, it makes the action easier to develop and test.

1. Test the action locally
   1. Use Node.js v20
   2. `cd packages/github-actions`
   3. `npm install`
   4. `node --watch actions/get-plugin-releases/src/get-plugin-releases.js`
   5. There should be no output because it's changed to only perform directly when running in GitHub Actions
   6. Edit the `packages/github-actions/actions/get-plugin-releases/src/get-plugin-releases.js` file and save it with the following test code at the end of the file
      ```js
      getPluginReleases( {
      	slug: 'wordpress',
      	numberOfReleases: 10,
      	includeRC: false,
      	includePatches: false,
      } );
      
      getPluginReleases( {
      	slug: 'woocommerce',
      	numberOfReleases: 10,
      	includeRC: true,
      	includePatches: true,
      } );
      ```
   7. There should be corresponding outputs
      image
2. View a previous workflow run used v1 action
   - https://github.com/woocommerce/google-listings-and-ads/actions/runs/8652279994
   - There are warnings of using Node.js 16
      image
3. View a test workflow run used updated action
   - https://github.com/eason9487/grow/actions/runs/8828174294
   - There's no more warning of using Node.js 16
      image
4. View the versions output
   - https://github.com/eason9487/grow/actions/runs/8828174294/job/24236656175#step:8:1
      image

### DIFF
--- a/packages/github-actions/actions/get-plugin-releases/README.md
+++ b/packages/github-actions/actions/get-plugin-releases/README.md
@@ -23,25 +23,25 @@ jobs:
     steps:
       - name: Get Release versions from WooCommerce
         id: wc-versions
-        uses: woocommerce/grow/get-plugin-releases@actions-v1
+        uses: woocommerce/grow/get-plugin-releases@actions-v2
         with:
           slug: woocommerce
 
       - name: Get Release versions from WordPress
         id: wp-versions
-        uses: woocommerce/grow/get-plugin-releases@actions-v1
+        uses: woocommerce/grow/get-plugin-releases@actions-v2
         with:
           slug: wordpress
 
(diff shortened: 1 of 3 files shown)


mikkamp (approved):
Thanks for the changes, they look good to me.

I can confirm I'm getting the correct results locally. Now that we use `fetch` directly I also don't see any flakey results being returned. The results from a previous run and a new run look good, and I can confirm the warnings are no longer present.

PR #122, opened 2024-04-26
Upgrade the `branch-label` action to use Node.js v20

eason9487:

Part of #108

This PR upgrades the `branch-label` action to use Node.js v20.
- The `actions-ecosystem/action-add-labels` action has not been updated for almost 4 years. Therefore, this PR changes to use an alternative `actions/labeler` action to achieve the same function.
- In addition, to avoid the issue of not having permission to add labels to pull requests, the example changes the recommended event to the `pull_request_target` instead of `pull_request`.

#### 📌 Using Node.js v20

1. View a previous workflow run used v1 action
   - https://github.com/woocommerce/google-listings-and-ads/actions/runs/8826963398
   - There are warnings of using Node.js 16 or even Node.js 12
      image
2. View test workflow runs used updated action
   - https://github.com/eason9487/grow/actions/runs/8845076148
   - There's no more warning of using Node.js 16
      image

#### 1️⃣ Add a label: https://github.com/eason9487/grow/pull/24

image

#### 2️⃣ Add multiple labels: https://github.com/eason9487/grow/pull/25

image

#### 3️⃣ Mismatched branch name: https://github.com/eason9487/grow/pull/26

No labels are added because the branch name `fixed/space-time-rift` didn't match any rules.

image

#### 4️⃣ No required permissions: https://github.com/eason9487/grow/pull/29

This test explains that [a workflow uses `pull_request` event](https://github.com/eason9487/grow/blob/fbfea2d2d8946ba72356ab8c71e7a9d92fc6f4d5/.github/workflows/test-github-actions-branch-label.yml#L3-L10) may result in permission issue when a PR is created from forks.

image
image

### DIFF
--- a/packages/github-actions/actions/branch-label/README.md
+++ b/packages/github-actions/actions/branch-label/README.md
@@ -17,12 +17,43 @@ See [action.yml](action.yml)
 
 ```yaml
 on:
-  pull_request:
+  pull_request_target:
     types: opened
+
 jobs:
   SetLabels:
+    permissions:
+      contents: read
+      pull-requests: write
     runs-on: ubuntu-latest
     steps:
       - name: Set Labels
-        uses: woocommerce/grow/branch-label@actions-v1
+        uses: woocommerce/grow/branch-label@actions-v2
+```
+
+#### Permissions:
+
+It's recommended to use the `pull_request_target` event instead of `pull_request` to avoid the issue of not having permission to add labels to pull requests.
+
+Ref:
+- https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#pull_request_target
+- https://github.com/actions/labeler/tree/v5#permissions
+
+## Migration from v1 to v2:
+
+```diff
+-  pull_request:
++  pull_request_target:
+     types: opened
+
+ jobs:
+   SetLabels:
++    permissions:
++      contents: read
++      pull-requests: write
+     runs-on: ubuntu-latest
+     steps:
+       - name: Set Labels
+-        uses: woocommerce/grow/branch-label@actions-v1
++        uses: woocommerce/grow/branch-label@actions-v2
 ```
--- a/packages/github-actions/actions/branch-label/README.md
+++ b/packages/github-actions/actions/branch-label/README.md
@@ -33,10 +33,11 @@ jobs:
 
 #### Permissions:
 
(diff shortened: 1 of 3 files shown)


packages/github-actions/actions/branch-label/README.md
```diff
@@ -17,12 +17,43 @@ See [action.yml](action.yml)
...
     runs-on: ubuntu-latest
     steps:
       - name: Set Labels
-        uses: woocommerce/grow/branch-label@actions-v1
+        uses: woocommerce/grow/branch-label@actions-v2
+```
+
+#### Permissions:
+
+It's recommended to use the `pull_request_target` event instead of `pull_request` to avoid the issue of not having permission to add labels to pull requests.
```
jorgemd24: In this action, we're not directly interacting with the code from the PR, so we're OK. However, shouldn't we also mention here that `pull_request_target` will grant access to secrets for PRs from forks?

packages/github-actions/actions/branch-label/README.md
```diff
@@ -17,12 +17,43 @@ See [action.yml](action.yml)
...
+
+Ref:
+- https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#pull_request_target
+- https://github.com/actions/labeler/tree/v5#permissions
```
jorgemd24, line 41: As we are  using the labeler action, I think this link can be useful too: https://github.com/actions/labeler?tab=readme-ov-file#usage

jorgemd24 (approved):
Thanks @eason9487 for working on this! I tested creating some PRs in a fork branch and it worked well, so LGTM 👍  I left some comments but they are not blockers.

packages/github-actions/actions/branch-label/README.md
eason9487: Thanks for the suggestion. Added in https://github.com/woocommerce/grow/pull/122/commits/424eada2c2806c5635d1467e2fa494e9ba2a992c.

packages/github-actions/actions/branch-label/README.md
eason9487: Added in https://github.com/woocommerce/grow/pull/122/commits/424eada2c2806c5635d1467e2fa494e9ba2a992c.

PR #123, opened 2024-04-26
Update the README for the `github-actions` package to be based on Node.js v20

eason9487:

Part of #108

This PR updates the README for the `github-actions` package to be based on Node.js v20

Read the updated [README](https://github.com/woocommerce/grow/blob/9b2bb6b42188c7725b7fe054c61f56d76f2879ff/packages/github-actions/README.md) to see if it is clear enough.

### DIFF
--- a/packages/github-actions/README.md
+++ b/packages/github-actions/README.md
@@ -6,6 +6,8 @@ Custom GitHub actions that help to composite GitHub workflows across the repos m
 
 ## Actions list
 
+All actions involving Node.js are run in v20.
+
 - [`automerge-released-trunk`](actions/automerge-released-trunk) - Merge `trunk` to `develop` after an extension release
 - [`branch-label`](actions/branch-label) - Set PR labels according to the branch name
 - [`coverage-report`](actions/coverage-report) - Add a clover coverage report as a PR comment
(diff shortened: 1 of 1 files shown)


packages/github-actions/README.md
```diff
@@ -25,11 +27,11 @@ Custom GitHub actions that help to composite GitHub workflows across the repos m
 - [`stylelint-annotation`](actions/stylelint-annotation) - Annotate stylelint results via stylelint formatter
 - [`update-version-tags`](actions/update-version-tags) - Update version tags
 
-## Prerequisites
+## Development
 
 ### JavaScript actions
 
-1. Install `node` with version >= 14
+1. Install `node` with version v20 (recommended) or v18 (compatible)
```
jorgemd24: Why not just go with v20? Otherwise, which specific version of v18 can I install or is compatible?

jorgemd24 (approved):
Hi @eason9487, thanks for updating the readme file. I left a comment about installing `v18 (compatible)`. Everything else looks good to me.

packages/github-actions/README.md
eason9487: Node.js v18 was listed because it's also compatible to use in development. Considering the End-of-Life date for v18 is within a year, it should be better to recommend v20 only. Thanks for the suggestion! Adjusted in https://github.com/woocommerce/grow/pull/123/commits/c97deef5de7472ef7e61bca22cd3a4b2a3b3e24a.

